PR #14, opened 2025-06-09
SOF-203: Replace Column with LazyColumn in LandingScreen for horizontal scrolling

ThomasT0830:
This pull request addresses the issue where users cannot access all the buttons on the LandingScreen when the device is held horizontally. By replacing the Column component in the LandingScreen with a LazyColumn, this update enables vertical scrolling on the LandingScreen, allowing users to access all buttons, even when the device is held horizontally, and ensures that no UI elements are cut off. The change was tested on both the emulator and a physical device.

https://github.com/user-attachments/assets/5662f7c3-49e3-40c9-9f72-0a4d821d4c8c

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingScreen.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingScreen.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.material3.Badge
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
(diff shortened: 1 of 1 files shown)


aryamanshodhan (changes requested):
This is a well-written PR — I really appreciate the cleanup and the optimizations you've introduced. Great work overall!

A couple of suggestions to consider:

1. Avoid wildcard imports: Using statements like `import androidx.compose.material3.*` is generally discouraged because they can lead to namespace pollution and make it harder to identify where specific symbols are coming from. It's better to import only the components you need explicitly — this improves code readability and makes automated tooling (like code search or auto-import) more precise. This is especially important given that our codebase is public and intended to be open source.

2. Consider `Column` with `verticalScroll` for static content: Currently, a `LazyColumn` is being used to render a static list of four buttons. I'd say `LazyColumn` is great for long or dynamic lists due to its performance benefits. However, since the content is fixed and relatively small, I recommend switching to a `Column` with `val verticalScrollState = rememberScrollState()` and `Modifier.verticalScroll(verticalScrollState)`. You can refer to how this is handled in `SurveillanceFormScreen.kt` for a similar use case.

Let me know if you’d like help with the changes — happy to pair or review a follow-up!.

ThomasT0830:
Thank you for the suggestions! I have implemented both changes, and the code should be ready for review again.

aryamanshodhan (approved):
LGTM

JohnnieShen (approved):

PR #15, opened 2025-06-10
SOF-207: Create WorkManager classes for Image and Session metadata upload.

aryamanshodhan:
This PR introduces the core infrastructure for handling background uploads using WorkManager. Specifically, it creates two separate CoroutineWorker classes: ImageUploadWorker and MetadataUploadWorker, which are responsible for uploading session image files and associated session metadata, respectively. These workers are enqueued as part of a sequential WorkManager chain to ensure metadata is uploaded prior to image data, maintaining logical consistency in the upload pipeline.

The current implementation simulates the upload process by introducing artificial delays for each image, allowing us to validate notification behavior, lifecycle persistence, and retry logic under constrained conditions (e.g., interrupted network). Each worker runs as a foreground service with its own notification channel to keep the user informed during upload progress, ensuring Android's background execution limits are respected.

This initial version does not yet interact with the backend but lays the groundwork for integrating actual network calls. Testing was performed manually by triggering the work sequence via the UI, observing notification updates, successful completion states, and automatic retry behavior when simulating failures (e.g., disabling Wi-Fi mid-upload). Logs were also inspected to verify correct work sequencing and constraint enforcement.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -130,6 +130,9 @@ dependencies {
     // Proto Data Store Library
     implementation(libs.androidx.datastore)
 
+    // Work Manager Library
+    implementation(libs.androidx.work.runtime.ktx)
+
     // JSON Serialization Library
     implementation(libs.kotlinx.serialization.json) // Kotlinx JSON serialization library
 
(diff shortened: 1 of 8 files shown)


ThomasT0830 (approved):
Everything looks great!

JohnnieShen (approved):
Looks good to me

PR #19, opened 2025-06-17
SOF-206: Create a new ProtoDataStore object for a unique device.

aryamanshodhan:
This pull request introduces a new `ProtoDataStore` implementation for persisting a unique device object locally using `kotlinx.serialization`. Rather than relying on a `.proto` schema, the device metadata is stored via a `DeviceDto` data class with custom serializers.

A domain-level `DeviceCache` interface has been defined, along with its implementation, `DeviceCacheImplementation`, which handles the conversion between the domain model and the serialized DTO. The cache persists key fields such as the device ID, model, registration timestamp, and associated program ID.

To support dependency injection, a Hilt module provides the configured `DataStore<DeviceDto>` instance. Logging was added to test and confirm that the device and program ID are correctly stored and retrieved, ensuring reliable state persistence across app sessions. This foundation will support device-level behavior such as tracking, analytics, and session attribution in future features.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/cache/DeviceCacheImplementation.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/cache/DeviceCacheImplementation.kt
@@ -1,0 +1,30 @@
+package com.vci.vectorcamapp.core.data.cache
+
+import androidx.datastore.core.DataStore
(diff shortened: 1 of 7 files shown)


ThomasT0830 (approved):
Everything looks good! Tested the code on a physical device

JohnnieShen (approved):

PR #21, opened 2025-06-18
SOF-129: GPS for Geolocation tracking of Latitude and Longitude

JohnnieShen:
This pull request added features related to location fetching, displaying and error handling for the `SurveillanceFormScreen`, A new `LocationClient` provides a one-shot getCurrentLocation() that, because we don't need a high accuracy location result, first checks lastLocation (if under 5 minutes old, to save battery) and otherwise falls back to a cancellable fresh fix via FusedLocationProviderClient.getCurrentLocation(...). Under the hood it uses suspendCancellableCoroutine and a CancellationTokenSource so that GPS requests are automatically torn down when the calling coroutine is cancelled. 

In the presentation layer, we introduce a sealed `LocationState` (Loading, Success(lat, lon), Error(message)) that is currently embedded in `SurveillanceFormState` but can be taken out later. The `SurveillanceFormViewModel`’s getLocation() launches on start, attempts up to two tries with a 5 second timeout (withTimeout(5_000)) and then updates locationState accordingly. 

This have been tested mostly on the physical device as emulators have their quirks and sometimes won't work with GPS emulation. More testing will be needed to mock GPS errors on the physical device to see if error handling is working correctly.

EDIT: Issue resolved regarding location data timing out on devices without any internet connection. Solved by switching from  PRIORITY_BALANCED_POWER_ACCURACY to  PRIORITY_HIGH_ACCURACY, with the latter prioritizing GPS data over cell+WIFI, which is what we need since the APP will be deployed in regions without internet connection, and the trade off with battery life is worth it. Also increased timeout limit from 5s to 30s because GPS cold start takes significantly longer and this number is up to debate.


ThomasT0830:
The code looks great! I did some testing on my device, and it seemed to have a bit of trouble fetching the location. It tried to get it but eventually showed 'Could not get location: GPS timeout.' It might just be something with my device, but I figured I would mention it just in case! Thanks!

JohnnieShen:
> The code looks great! I did some testing on my device, and it seemed to have a bit of trouble fetching the location. It tried to get it but eventually showed 'Could not get location: GPS timeout.' It might just be something with my device, but I figured I would mention it just in case! Thanks!

Thank you for catching that! I've tried a few times on my device and it worked for me. Could you quickly double check and see if the device is set up correctly? e.g. if location is enabled and permission is granted correctly. Not sure what could've caused the timeout so just want to quickly run a couple causes by you.

ThomasT0830:
Absolutely! The location is currently turned on, and the app is given the "Only while app is in use" permission. I will try to test it out a bit more to see if I can understand what is going on. Hope that helps, and thanks so much!

JohnnieShen:
> Absolutely! The location is currently turned on, and the app is given the "Only while app is in use" permission. I will try to test it out a bit more to see if I can understand what is going on. Hope that helps, and thanks so much!

Ty! If it still persists, and if it works for you we could swap devices for a day or two just so we can both make sure our code works on different hardware.

ThomasT0830:
Yeah, that sounds like a great idea!

ThomasT0830 (dismissed):
Everything looks good to me! It seems that the physical device was able to fetch the GPS location when the internet connection was enabled. However, when the internet was turned off, GPS location fetching timed out.

JohnnieShen:
Still timing out? Might be that there are no GPS signal in your room? My desk was in front of a window and could have better coverage.

app/src/main/java/com/vci/vectorcamapp/location/data/LocationClient.kt
aryamanshodhan, line 1: Couple of suggestions:
1. Highly suggest removing SuppressLint, and handling the case for when location permission is not granted. Might be bulky, but I would not suggest suppressing warnings/errors.
2. Maybe you could provide the fusedLocationProviderClient using Hilt in a separate module under the surveillance_form package?
3. I would also suggest just moving all these location related files into the surveillance_form module, because the top level folders indicate features or screens, and by itself location is not really an app feature. It is a feature of another module (in this case the surveillance_form module). Eg. for Room, we have a separate module under core. Similarly define data/location, and domain/location within surveillance_form.
4. Can you explain what MAX_LAST_LOCATION_AGE_MS is?

app/src/main/java/com/vci/vectorcamapp/location/data/repository/LocationRepositoryImplementation.kt
aryamanshodhan, line 1: No need to call this @Singleton if you are already marking it in the LocationModule.

app/src/main/java/com/vci/vectorcamapp/location/di/LocationModule.kt
aryamanshodhan, line 1: Since this will be a feature of the form, scope to the view model. Not the entire app.
`@InstallIn(ViewModelScoped::class)`

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormState.kt
aryamanshodhan, line 1: Your definition of LocationState looks very similar to the Result class we have defined. If want to check Results from execution of functions, I would suggest using that because it provides a way to define custom error messages. The Result class also provides a couple of neat functions like Result.onSuccess and Result.onFailure. I would suggest using that instead.

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormViewModel.kt
aryamanshodhan, line 1: Can you explain what the Repeat(2) is doing?

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
aryamanshodhan, line 1: Define 2 state variables called latitude and longitude which you then display on the screen of the app instead of logging to the console. Revisit the use of LocationState. If you need to wait until things are loaded, set the isLoading state variable to true. For Success and Error situations, refer to the Result class in the core.domain.util module.

aryamanshodhan (changes requested):
Well done Johnny! This is written well overall. I have requested some changes for this PR. Thanks!

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormViewModel.kt
JohnnieShen: it re-attempts to fetch location for up to 2 more times, in case the first few times failed due to some circumstances, if you think it's not necessary i can remove it, but i think it is good to have

aryamanshodhan: Yeah we can do that. Instead of Repeat(2), can you define a const variable called `MAX_ATTEMPTS = 2`? Just to minimize hard coding

app/src/main/java/com/vci/vectorcamapp/location/data/LocationClient.kt
JohnnieShen: MAX_LAST_LOCATION_AGE_MS is used when fetching the last known location of the device. This is a workaround I came around for lowering battery consumption since we are currently using high accuracy mode. Currently the device will first attempt to get the last known location of the device (which is passive and so requires less power than actively requesting a new location), and then compares it to MAX_LAST_LOCATION_AGE_MS, which is a threshold for how old the last known location is, if it is older than MAX_LAST_LOCATION_AGE_MS (currently set to 5 minutes), it would request a new location.

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
JohnnieShen: Latitude and longtitude are already displayed on the screen, between note and the submit button

JohnnieShen: Should I still have 2 state variables called latitude and longitude if I am to use the Result class and have a pair for Success? I am wondering which solution would have better programming pattern, since Result have good functionalities built in but exposing latitude and longitude instead of a tuple is clearer). 

Additionally, for this case I don't believe setting isLoading to true when waiting for GPS signal to load is necessary because it would block the entire screen and the user won't be able to fill in other fields during then. An idea I have right now is to gray out the submit button until GPS either loads a location or times out. Do you think this would work?

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/LocationClient.kt
```diff
@@ -0,0 +1,78 @@
...
+class LocationClient @Inject constructor(
+    private val fused: FusedLocationProviderClient,
+    @ApplicationContext private val context: Context
+) {
+    companion object {
+        private const val MAX_LAST_LOCATION_AGE_MS = 5 * 60 * 1_000L
+    }
+
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    suspend fun getCurrentLocation(): Location = suspendCancellableCoroutine { cont ->
```
aryamanshodhan, line 30: Is there a solution rather than using @RequiresPermission?

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/FusedLocationProviderModule.kt
```diff
@@ -0,0 +1,22 @@
...
+@Module
+@InstallIn(ViewModelComponent::class)
+object FusedLocationProviderModule {
+    @Provides
+    @ViewModelScoped
+    fun provideFusedLocationProviderClient(
+        @ApplicationContext context: Context
+    ): FusedLocationProviderClient =
+        LocationServices.getFusedLocationProviderClient(context)
+}
```
aryamanshodhan: End all files with an empty line

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/LocationErrorToString.kt
```diff
@@ -0,0 +1,13 @@
+package com.vci.vectorcamapp.surveillance_form.location.data
+
+import android.content.Context
+import com.vci.vectorcamapp.R
+
+fun LocationError.toString(context: Context): String {
+    val resId = when(this) {
+        LocationError.PERMISSION_DENIED -> R.string.location_permission_error
+        LocationError.GPS_TIMEOUT -> R.string.location_gps_timeout
```
aryamanshodhan: Sorry for being so pedantic, but to align the naming convention with other errors, please rename this to:
1. R.string.location_error_permission_denied
2. R.string.location_error_gps_timeout
3. R.string.location_error_unknown

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/repository/LocationRepositoryImplementation.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.vci.vectorcamapp.surveillance_form.location.data.LocationClient
+import com.vci.vectorcamapp.surveillance_form.location.domain.repository.LocationRepository
+import javax.inject.Inject
+import javax.inject.Singleton
+
+class LocationRepositoryImplementation @Inject constructor(
+    private val locationClient: LocationClient
+) : LocationRepository {
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    override suspend fun getCurrentLocation(): Location =
```
aryamanshodhan, line 11: Is there another option rather than using @RequiresPermission?

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/di/LocationModule.kt
```diff
@@ -0,0 +1,20 @@
...
+
+@Module
+@InstallIn(ViewModelComponent::class)
+abstract class LocationModule {
+
+    @Binds
+    @ViewModelScoped
+    abstract fun bindLocationRepository(
+        impl: LocationRepositoryImplementation
+    ): LocationRepository
```
aryamanshodhan, line 19: locationRepositoryImplementation: LocationRepositoryImplementation

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
```diff
@@ -32,6 +37,8 @@ fun SurveillanceFormScreen(
     onAction: (SurveillanceFormAction) -> Unit,
     modifier: Modifier = Modifier
 ) {
+    val ctx = LocalContext.current
```
aryamanshodhan: Use full names where possible. context, latitude, longitude

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/LocationClient.kt
```diff
@@ -0,0 +1,78 @@
...
+                context, Manifest.permission.ACCESS_FINE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED
+        ) {
+            cont.resumeWithException(
+                SecurityException("ACCESS_FINE_LOCATION permission not granted")
+            )
+            return@suspendCancellableCoroutine
+        }
+
+            val cts = CancellationTokenSource()
```
aryamanshodhan: Use full names where possible. continuation, cancellationTokenSource

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
```diff
@@ -195,6 +202,21 @@ fun SurveillanceFormScreen(
             onValueChange = { onAction(SurveillanceFormAction.EnterNotes(it)) },
             modifier = modifier
         )
+        state.locationResult?.let { result ->
+
+            result.onSuccess { (lat, lon) ->
+                Text("Latitude: $lat", Modifier.padding(vertical = 4.dp))
+                Text("Longitude: $lon", Modifier.padding(vertical = 4.dp))
+            }
```
aryamanshodhan: You can also just do:

result.onSuccess {
     ...
}.onError {
     ...
}

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormState.kt
```diff
@@ -3,11 +3,17 @@ package com.vci.vectorcamapp.surveillance_form.presentation
...
+import com.vci.vectorcamapp.core.domain.util.Result
+
 import java.util.UUID
 
 data class SurveillanceFormState(
     val isLoading: Boolean = false,
+    val locationResult: Result<Pair<Float,Float>,LocationError>? = null,
+    val latitude: Float? = null,
+    val longitude: Float? = null,
     val allSitesInProgram: List<Site> = emptyList(),
```
aryamanshodhan, line 17: I think you should store either the entire result, or the latitude and longitude, and error message. I would prefer the latter. Whenever you get the result, just update the state.

```kotlin
locationResult.onSuccess { (latitude, longitude) ->
     _state.update {
          it.copy(latitude = latitude, longitude = longitude)
     }
}.onError { error ->
     _state.update {
          it.copy(error = error)
     }
}
```

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormViewModel.kt
```diff
@@ -361,13 +372,29 @@ class SurveillanceFormViewModel @Inject constructor(
...
-        viewModelScope.launch {
-            // FETCH LOCATION DATA HERE
-            _state.update {
-                it.copy(
-                    isLoading = false
-                )
+    fun getLocation() = viewModelScope.launch {
+        _state.update { it.copy(isLoading = false) }
+        _state.update { it.copy(locationResult = null) }
+
```
aryamanshodhan: Replace the result with the lat and long, and error.

app/src/main/res/values/strings.xml
```diff
@@ -102,5 +102,13 @@
...
-
+  
+    <!-- Location error: Permission error -->
+    <string name="location_permission_error">Permission error. Please enable location access in settings.</string>
+  
+    <!-- Location error: GPS timeout error -->
+    <string name="location_gps_timeout">GPS Timeout. Please try again</string>
+  
+    <!-- Location error: Unknown error -->
+    <string name="location_unknown">An unknown error occurred. Please try again or contact support if the issue persists.</string>
```
aryamanshodhan: Names of the string resources should be consistent with the others.

aryamanshodhan (changes requested):

aryamanshodhan:
Well done Johnny! This is written well. There are some very small naming related suggestions I have made. But otherwise, this is written very well!

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/repository/LocationRepositoryImplementation.kt
JohnnieShen: I'm not sure I understand what you meant. Since the missing permission case is handled below explicitly, the require permission check here is just for getting rid of the lint warning. I can take out the permission managing logic into a seperate manager to decouple logic but that would still need a RequirePermission or SuppressLint flag here.

aryamanshodhan: I guess what I meant is, can you programmatically check to ensure that those permissions are accepted by the user? I feel like using the @RequiresPermission just to suppress the linter might not really be solving the underlying reason why the linter is complaining in the first place.

JohnnieShen: I can add another check for permission that will re-request if the permission is not given (even though it is unlikely to happen as we have a onResume in main activity that re checks for permission every time the user returns back to the app). However, I do believe that to resolve the linter warning, either RequirePermission or SuppressLint is necessary here, and keeping the RequirePermission annotation is ideal.

app/src/main/java/com/vci/vectorcamapp/surveillance_form/location/data/LocationClient.kt
```diff
@@ -0,0 +1,87 @@
...
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.android.scopes.ViewModelScoped
+import kotlinx.coroutines.CancellableContinuation
+import kotlinx.coroutines.suspendCancellableCoroutine
+import javax.inject.Inject
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+
+@ViewModelScoped
```
aryamanshodhan: You don't need @ViewModelScoped here.

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormAction.kt
```diff
@@ -26,4 +26,5 @@ sealed interface SurveillanceFormAction {
     data class SelectCollectionMethod(val option: CollectionMethodOption) : SurveillanceFormAction
     data class SelectSpecimenCondition(val option: SpecimenConditionOption) : SurveillanceFormAction
     data class EnterNotes(val text: String) : SurveillanceFormAction
+    class RetryLocation: SurveillanceFormAction
```
aryamanshodhan: data object...not class.

app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormViewModel.kt
```diff
@@ -388,6 +403,34 @@ class SurveillanceFormViewModel @Inject constructor(
                     selectedSentinelSite = sentinelSite
                 )
             }
+
+            getLocation()
+        }
+    }
+
+    suspend fun getLocation() {
+        repeat(MAX_ATTEMPTS) { attempt ->
```
aryamanshodhan, line 412: make this a private function if you arent exposing it elsewhere.

aryamanshodhan (changes requested):
Well done Johnny! Very minor changes needed here. Otherwise everything looks good!

ThomasT0830 (changes requested):
Unfortunately, I am currently getting a compilation error: 


Expression 'RetryLocation' of type 'SurveillanceFormAction.RetryLocation' cannot be invoked as a function. The function 'invoke()' is not found

aryamanshodhan (dismissed):
Good catch! Me too!

aryamanshodhan (approved):

aryamanshodhan (changes requested):
Error on the app

ThomasT0830 (approved):
Everything looks good to me! Tested it on a physical device

aryamanshodhan (approved):
Looks good. One thing I will say though, the lat and long values updated immediately even though I did not have wifi.

PR #23, opened 2025-06-20
SOF-183: Fixed bounding box UI updates issue after imaging button is clicked

ThomasT0830:
This pull request fixes an issue on the Imaging screen where the bounding box appears in the wrong position if the phone is rotated immediately after the imaging button is clicked. To resolve this issue, I added a previewBoundingBoxUi property in the ImagingState. This separates the bounding box used for the live preview from the one associated with the actual capture. To improve clarity, I also renamed the original currentBoundingBoxUi to captureBoundingBoxUi. The code was tested on a physical device.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
@@ -131,9 +131,10 @@ class ImagingViewModel @Inject constructor(
                         _state.update {
                             it.copy(
                                 currentSpecimen = it.currentSpecimen.copy(id = specimenId),
-                                currentBoundingBoxUi = boundingBox?.let { boundingBox ->
-                                    inferenceRepository.convertToBoundingBoxUi(boundingBox)
-                                })
+                                previewBoundingBoxUi = boundingBox?.let { box ->
+                                    inferenceRepository.convertToBoundingBoxUi(box)
+                                }
+                            )
                         }
                     } catch (e: Exception) {
                         Log.e("ViewModel", "Image processing setup failed: ${e.message}")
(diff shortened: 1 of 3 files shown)


JohnnieShen (dismissed):
Looks good to me!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -210,7 +211,10 @@ class ImagingViewModel @Inject constructor(
                                         species = species.label,
                                         sex = sex?.label,
                                         abdomenStatus = abdomenStatus?.label,
-                                    ), currentImage = bitmap
+                                    ),
+                                    currentImage = bitmap,
+                                    captureBoundingBoxUi = it.previewBoundingBoxUi,
```
aryamanshodhan, line 218: Unfortunately this did not work for me, especially if I increase the frame rate in `SpecimenImageAnalyzer`. If you click the capture button and immediately move the device to the side, the issue still persists> I think you will need to extract the detection logic out of the `inferenceRepository.classify` function. The bounding box which you get during that frame should be set to the `captureBoundingBoxUi`.

aryamanshodhan (changes requested):
Unfortunately, I am still experiencing the issue...check out the comment I left.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
aryamanshodhan: Actually a more elegant solution which might work is to only update the previewBoundingBox in processFrame if the captureBoundingBox is null. Because if that it not the case, it means the image has already been captured. Then you can keep:
captureBoundingBoxUi = it.previewBoundingBoxUi
previewBoundingBoxUi = null
on Lines 216 and 217

aryamanshodhan (approved):
Well done! This worked for me. I hope you tested on your end as well, because your phones update the bounding box faster than mine so it might be more likely to cause issues on your end.

PR #24, opened 2025-06-20
SOF-214: Create an initial registration screen which opens the first time the app is opened

JohnnieShen:
This PR creates a new screen for registration when the app is first opened for the goal of allowing the user to select which program this phone belongs to and keep them from changing in the future. 

The implementation includes a new screen called RegistrationScreen with a drop down menu on it, the user will be able to select a program and the id will be stored in the local DataStore. The app will check for whether a program ID is available, if it is available, it means that the device has registered already and the user will be lead to landing screen, if not, the user will be lead to registration screen, where it would prompt the user to select which program the device would belong to. During registration, it would also create a new Device data object and store it locally.

Another change is that SurveillanceFormDropdownOptions interface have been renamed to DropdownOption, because it is now no longer used exclusively in surveillance form. 

Testing have been performed and bugs have been fixed, more testing on different devices is welcomed.


app/src/main/java/com/vci/vectorcamapp/registration/presentation/RegistrationAction.kt
aryamanshodhan, line 1: Think of a more descriptive name instead of Continue. Perhaps something like "ConfirmRegistration" or "SignUpDevice" or "RegisterDevice".

app/src/main/java/com/vci/vectorcamapp/registration/presentation/RegistrationEvent.kt
aryamanshodhan, line 1: Rename to `NavigateToLandingScreen` to be more consistent with other files.

app/src/main/java/com/vci/vectorcamapp/core/domain/repository/ProgramRepository.kt
aryamanshodhan, line 1: I think you can make this a suspend function and not use flows, because data is not changing so you arent really observing anything.

app/src/main/java/com/vci/vectorcamapp/MainActivity.kt
aryamanshodhan, line 1: A much cleaner way in which you can do this is rename the PermissionViewModel to something related to the main activity (it basically would store the state for the whole MainActivity). Then you could have a state variable for the selected program ID. That way you dont even need the observeProgramId function. You can just keep track of the selected programId in the `MainState` if you will. Also you can move all the logic associated with determining the start destination to a state variable which you can just pass into the nav graph using state.startDestination.

app/src/main/java/com/vci/vectorcamapp/core/data/cache/DeviceCacheImplementation.kt
aryamanshodhan, line 1: This is probably not needed because the programId is not being observed. It is just being used to determine what the start destination should be. Again, if that logic is moved to a view model, you dont need to do this.

app/src/main/java/com/vci/vectorcamapp/core/data/repository/ProgramRepositoryImplementation.kt
aryamanshodhan, line 1: For simple display purposes, you dont need a flow. You are better off with just a suspend function which returns a lis of programs

app/src/main/java/com/vci/vectorcamapp/core/domain/cache/DeviceCache.kt
aryamanshodhan, line 1: I think this function is not needed.

app/src/main/java/com/vci/vectorcamapp/navigation/NavGraph.kt
aryamanshodhan, line 1: Make the naming consistent as in the other parts of this file. vm should be replaced with viewModel. Also, can you explain the purpose of:

```kotlin
navController.navigate(Destination.Landing) {
                        popUpTo(Destination.Registration) { inclusive = true }
                    }
```

Why not just `navController.navigate(Destination.Landing)`

app/src/main/java/com/vci/vectorcamapp/registration/presentation/RegistrationScreen.kt
aryamanshodhan, line 1: Separate composables in separate files. Move ProgramItem to a separate file under a new package called components/. Directly use:

DropdownField(
     ...
     options = state.programs.map { ProgramOption(it.id, it.name) }
     ...
)

app/src/main/java/com/vci/vectorcamapp/registration/domain/enums/ProgramOption.kt
aryamanshodhan, line 1: Not sure if you need an id field. Check out how site and districts are implemented in the session form

app/src/main/java/com/vci/vectorcamapp/registration/presentation/RegistrationState.kt
aryamanshodhan, line 1: Shouldn't selected program be a string because that is what the value should be when you select the name of the program right? Check out how this is implemented for the sites and districts in the surveillance form.

app/src/main/java/com/vci/vectorcamapp/registration/presentation/RegistrationViewModel.kt
aryamanshodhan, line 1: You dont need Channel.BUFFERED. Consider loading all programs in the same way in which sites and districts are loaded in the surveillance form screen. Check it out. That avoids the need for the init block. Why do you need a dummy device? The model can be got using:
`Build.MANUFACTURER + " " + Build.MODEL`
Since we are not sending the device to the backend yet, registeredAt should be 0L, and the deviceId should be -1. Only when we submit the device to the backend, we should overwrite the ID and registeredAt times with what the backend returns.

aryamanshodhan (changes requested):
Good job Johnny! This is written very well. There are still some opportunities to significantly make the code cleaner. Please look through my commets, and make the changes ASAP so we can get everything merged.

app/src/main/java/com/vci/vectorcamapp/core/domain/repository/ProgramRepository.kt
JohnnieShen: I was using a flow mainly because in the future we might need to dynamically fetch all available programs and a flow would be easier for avoiding race conditions and making sure the data updates correctly.

aryamanshodhan:
Additionally, please find ALL instances where the programID has been hard coded, and replace it with the selected programId.

app/src/main/java/com/vci/vectorcamapp/navigation/NavGraph.kt
JohnnieShen: The difference is if the user moves from Registration to Landing, if `popUpTo(Destination.Registration) { inclusive = true }` is executed then Registration will be popped from the stack and therefore if the user press back they will not be lead back to Registration, while if we use `navController.navigate(Destination.Landing)` the user will be able to return to Registration manually which is not ideal

app/src/main/java/com/vci/vectorcamapp/core/domain/repository/ProgramRepository.kt
aryamanshodhan: Can you give an example for such a race condition and when we might want to do that?

JohnnieShen: I couldn't really think of any right now and I think I might be overthinking this. In this case i agree a suspend function is better.

aryamanshodhan (commented):
I made the most recent changes to this branch. Thomas and Johnny, please look through my changes and let me know. Johnny, once I have your approval that this looks good, I will approve the PR from my end. Thomas you can approve or suggest changes the normal way.

aryamanshodhan (approved):
Approved by Johnny

ThomasT0830 (approved):
Everything looks good to me! Tested the code on a physical device.

PR #25, opened 2025-06-22
SOF-218: Modify SpecimenDetector to detect multiple specimens and disable capture when there are more than 2 specimens on screen.

aryamanshodhan:
This PR introduces validation to ensure images are only captured when exactly one specimen is detected in the frame, preventing captures with zero or multiple specimens.
The detection system now returns all detected specimens instead of just the best match. Capture logic now validates specimen count in ImagingViewModel.captureImage(). Images are only captured when exactly one specimen is detected, with appropriate error messages shown for zero or multiple specimens.
The UI displays multiple bounding boxes simultaneously in camera preview. State management changed from single previewBoundingBoxUi to previewBoundingBoxesUiList to support multiple detection results.
Error handling includes new MULTIPLE_SPECIMENS_FOUND error type with message "Multiple Specimens Detected. Please make sure exactly one specimen is in frame." Specimen classification workflow refactored to work with pre-cropped images while maintaining backward compatibility.
Testing was performed on device using images containing 0, 1, and >2 specimens to validate proper error handling, successful capture behavior, and correct bounding box rendering in camera preview.


JohnnieShen (approved):
Tested on a physical device. The detection of multiple specimens is not ideal, but the content of this PR worked as expected.

aryamanshodhan:
Yeah...that's unfortunately a YOLO issue because the model isn't trained on specimens with multiple images...so it doesn't do very well.

ThomasT0830 (approved):
There seems to be an issue with the model, but otherwise, everything looks good!

PR #29, opened 2025-06-27
SOF-201: Fix image rotation inconsistencies when the phone is rotated

ThomasT0830:
This pull request addresses a critical issue on the Imaging Screen where captured images appeared rotated or flipped when the device was rotated. Although the expected behavior is for all captured images to remain upright and in portrait orientation, rotating the phone sometimes caused images to appear sideways, upside down, or mirrored. 

The underlying problem was traced to the toUprightBitmap function in ImageProxyExtensions. Previously, this function corrected image orientation using only the image’s sensor rotation (imageInfo.rotationDegrees). However, this does not account for the actual orientation of the device display. As a result, the image could be misaligned with the display’s orientation, especially when the device was rotated between portrait and landscape modes. 

To resolve this, the toUprightBitmap function has been updated to consider both the image’s rotation and the current display rotation. This allows the function to calculate and apply a correct total rotation, ensuring that the resulting image is always presented upright and in portrait orientation, regardless of how the phone is held. 

The code was tested on a physical device.


app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -13,6 +15,8 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.asImageBitmap
+import androidx.compose.ui.graphics.painter.BitmapPainter
 import androidx.compose.ui.platform.LocalContext
```
aryamanshodhan: Unused imports need to be removed.

aryamanshodhan (changes requested):
Well done Thomas! Please address the comment for now. There was one issue where if the device is in landscape mode and the UI is landscape, the bounding box does not display correctly. Can you look into this after you finish some of your other tasks? Everything works in portrait mode which is the most important thing. If time permits after your other tasks, please look into this. Thanks!

aryamanshodhan (approved):

JohnnieShen (approved):
Tested, assume I did the replication correctly I didn't see any issues so LGTM.

PR #31, opened 2025-06-28
SOF-181: Standardized use of modifier and Modifier

JohnnieShen:
This pull request refactors the code and the usage of modifier and Modifier. Now the root container uses the injected modifier while internal components uses Modifier (capitalized). Tested on physical device by looking at the potentially affected screens. More testing welcomed.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
@@ -43,16 +43,16 @@ fun LiveCameraPreviewPage(
                     controller.bindToLifecycle(lifecycleOwner)
                 }
             },
-            modifier = modifier.fillMaxSize()
+            modifier = Modifier.fillMaxSize()
         )
 
         boundingBoxesUiList.map {
-            BoundingBoxOverlay(it, modifier.fillMaxSize())
+            BoundingBoxOverlay(it, Modifier.fillMaxSize())
         }
 
         IconButton(
             onClick = onSaveSessionProgress,
-            modifier = modifier
+            modifier = Modifier
                 .padding(24.dp)
                 .size(64.dp)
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -43,16 +43,16 @@ fun LiveCameraPreviewPage(
                     controller.bindToLifecycle(lifecycleOwner)
                 }
             },
-            modifier = modifier.fillMaxSize()
+            modifier = Modifier.fillMaxSize()
```
aryamanshodhan, line 46: Can you please explain the use of Modifier vs. modifier? The root level modifier should be Modifier (Eg. wherever the first component in the hierarchy is. For example, the NavGraph is where the screens are rendering for the first time. Is that correct? Wherever you need the modifier to be relative to the previous screen in the hierarchy, you want to use modifier. If it should start a new hierarchy of constraints, you should use Modifier. Do these changes of modifier to Modifier address this?

aryamanshodhan (changes requested):
Well done Johnny! Wondering what your thoughts are about the comment I left. I will approve these changes once you respond to the comment. Thanks!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
JohnnieShen: My idea is that root level containers should use modifier and inner components should use Modifier, because root level containers would need to take the modifier passed in from navgraph, but inner components should use their own modifier to make sure nothing is messed up. In this file only the outermost Box would use the passed in modifier. Do you think this is a good idea?

aryamanshodhan (approved):

ThomasT0830 (approved):
Looks great to me! Tested the code on a physical device

PR #32, opened 2025-06-30
SOF-233: Initial flicker fix

JohnnieShen:
This pull request fixed an issue with flickers when first launching the app. Now the code no longer sets loading to true after seeing that it would need to request permission, thus solving the loading animation appearing after the first screen issue.


JohnnieShen:
Consolidated code, added placeholder splash screen, fixed flicker when launching landing screen, now the entire workflow should be smooth without flickers.

aryamanshodhan (changes requested):
Hey Johnny! This is good, but did you review the video I sent on the slack channel? It will make your life a LOT easier. Can you implement it in the way shown in that video instead? Thanks!

Attaching the video here as well:
https://youtu.be/eFZmMSm1G1c?si=ww5o8JlNwSFlYHUa

aryamanshodhan (approved):

PR #34, opened 2025-07-02
SOF-130: Enable autofocus to centroid of YOLO box in the camera preview

JohnnieShen:
This PR introduces functionalities related to autofocus and manual focus in imaging preview screen. In the imaging screen, when there are specimens in the preview frame, when bounding boxes appear, the camera will be auto focused on the center of the bounding boxes. The user can also click on the camera frame and manually set focus points to override auto focus, which would be displayed as a cyan circle, clicking on the circle again would cancel the manual focus override.

This have been tested and taking images after setting focus points would result in images focusing on the right thing. 

This PR would break imaging screen in landscape mode, as the camera preview frame would fill the entire screen.


ThomasT0830 (changes requested):
Works well on the physical device! Could consider moving some of the logic out of the Composable

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -30,23 +48,83 @@ fun LiveCameraPreviewPage(
...
                 }
-            },
-            modifier = Modifier.fillMaxSize()
-        )
+            }
+        }
 
-        boundingBoxesUiList.map {
+        boundingBoxesUiList.forEach {
             BoundingBoxOverlay(it, Modifier.fillMaxSize())
```
aryamanshodhan, line 138: keep it map. I think map is a bit more efficient

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -98,4 +176,4 @@ fun LiveCameraPreviewPage(
             )
         }
     }
-}
+}
```
aryamanshodhan: end with newline

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/CameraPreviewUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        val initialX = focusPoint.x - (focusRingSizePx / 2)
+        val initialY = focusPoint.y - (focusRingSizePx / 2)
+
+        val clampedX = initialX.coerceIn(0f, containerWidthPx - focusRingSizePx)
+        val clampedY = initialY.coerceIn(0f, containerHeightPx - focusRingSizePx)
+
+        return Offset(clampedX.toDp().value, clampedY.toDp().value)
+    }
+}
```
aryamanshodhan: These should not be independent functions. Look into how these functions might be placed into a manager interface or created into use cases. Research a bit about what is the best place for these functions within the clean architecture design. The functions themselves are written well. Also look into whether it is good practice to pass PreviewViews and controllers around like this, especially when we do not have access to context.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -30,23 +48,83 @@ fun LiveCameraPreviewPage(
     onSubmitSession: () -> Unit,
     modifier: Modifier = Modifier
 ) {
+    val context = LocalContext.current
     val lifecycleOwner = LocalLifecycleOwner.current
+    val previewView = remember { PreviewView(context) }
+    val density = LocalDensity.current
```
aryamanshodhan, line 59: Is this good practice? Remembering views like this?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -30,23 +48,83 @@ fun LiveCameraPreviewPage(
...
+
```
aryamanshodhan, line 60: If you conclude that it is fine to remember views, I would suggest changing to:
remember(context) { PreviewView(context) }

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -30,23 +48,83 @@ fun LiveCameraPreviewPage(
...
+    var overlaySize by remember { mutableStateOf(IntSize.Zero) }
+
+    LaunchedEffect(Unit) {
+        previewView.apply {
+            scaleType = PreviewView.ScaleType.FIT_CENTER
+            this.controller = controller
+        }
+        controller.bindToLifecycle(lifecycleOwner)
+    }
+
```
aryamanshodhan, line 78: Can't you just apply these changes when you are creating the PreviewView(context) object? Why does this need to be in a separate LaunchedEffect() call? I personally find LaunchedEffect and DisposableEffect a bit messy, so unless you absolutely need them, I would recommend against using them.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -30,23 +48,83 @@ fun LiveCameraPreviewPage(
...
+    LaunchedEffect(boundingBoxesUiList, manualFocusPoint) {
+        if (manualFocusPoint == null) {
+            if (boundingBoxesUiList.isNotEmpty()) {
+                autoFocusOnObject(controller, previewView, boundingBoxesUiList.first())
+            } else {
+                cancelFocus(controller)
+            }
+        }
+    }
```
aryamanshodhan, line 88: I think you can manage the focus point using ViewModel and state

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -30,23 +48,83 @@ fun LiveCameraPreviewPage(
...
+            ) {
+                manualFocusPoint?.let { focusPoint ->
+                    if (overlaySize != IntSize.Zero) {
+                        val offset = calculateFocusRingOffset(
+                            focusPoint = focusPoint,
+                            overlaySize = overlaySize,
+                            focusBoxSize = focusBoxSize,
+                            density = density
+                        )
+
```
aryamanshodhan, line 121: A lot of this logic can be moved into the view model. All logic should be moved into the view model. Try to be as intentional as possible when trying to apply separation of concerns.

aryamanshodhan (changes requested):
Hey Johnny! This is good. However, there are some cases where the code does not adhere to clean architecture. Please do some research into what the optimal structure might be for implementing this functionality. There are a couple of things I question (Eg. remembering views, moving the controller around into the util functions, etc.). Please do some research (using chatgpt or otherwise) to see if these are good practices. You can ask chatGpt or an AI assistant to provide feedback on whether this code adheres to clean architecture principles. Also address my comments please. Thanks.

aryamanshodhan:
Couple of things to watch out for, from GPT:
1. Null Safety Violation
controller.cameraControl?.startFocusAndMetering(action)
While using safe call (?.), there's no handling of null cases. The operation silently fails if cameraControl is null.

2. Memory Leak Risk
.disableAutoCancel()
Disabling auto-cancel on focus actions without proper cleanup can lead to resource leaks, especially if the camera lifecycle ends unexpectedly.

3. Inconsistent Unit Handling
return Offset(clampedX.toDp().value, clampedY.toDp().value)
This converts pixels to DP values but returns them as raw floats in an Offset, which expects pixel coordinates. This will cause incorrect positioning.

4. Expensive Recomposition
LaunchedEffect(boundingBoxesUiList, manualFocusPoint) {
This triggers on every bounding box list change, even if the list contents are identical but the reference changed.

5. Tight Coupling
The composable directly handles camera controller logic instead of delegating to a ViewModel or use case.

6. Hardcoded Values
Magic numbers like 64.dp, 4f / 3f, and 2.dp should be extracted to constants or theme values.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
JohnnieShen: I believe that this step is essential and using a seperate LaunchedEffect is not a bad idea. Android recommends that composable should be side effects free, in this case since we are connecting an external system to the compose lifecycle, we have to run it in controlled environment, and I believe that LaunchedEffect should be used here. I can use some dirtier hacks to get around it but I think this is the best way of doing it.

ThomasT0830 (approved):

aryamanshodhan (approved):

PR #35, opened 2025-07-03
SOF-239: Location client crash if session saved when requestFreshLocation is running

JohnnieShen:
This PR fixes an issue where if the requestFreshLocation is still running if the session is saved the app crashes. An if (cont.isActive) { check is added to the FailureListener to make sure that the token is still active to request fresh location.


aryamanshodhan (approved):
Looks good!

ThomasT0830 (approved):
Everything looks good!

PR #36, opened 2025-07-04
SOF-193: UI - Landing Screen

aryamanshodhan:
This PR introduces the initial implementation of the LandingScreen UI, including a fixed header section with a green gradient background and a set of categorized action tiles for navigating to key app features. The layout is structured using a Box to overlay the header and scrollable content, and the tiles are organized under clearly labeled sections like "Imaging" and "Library" to improve user navigation and clarity.

The main content area is designed to scroll vertically so that users can access all tiles regardless of screen size. However, there is currently an issue where the cards scroll underneath the header instead of being properly clipped or constrained. The intended behavior is for the scrolling to start below the gradient header, and for the scrollable content to be bounded so it doesn’t visually overlap with the fixed top section.

I’ve tried using padding and offset to position the scrollable area correctly, but the behavior still isn’t quite right. If anyone has suggestions on how to properly constrain the scroll region or clip it below a fixed height (e.g., 25% of screen height), I’d really appreciate the help.


ThomasT0830 (approved):
Everything looks good to me!

JohnnieShen (approved):

PR #42, opened 2025-07-08
SOF-246: Fix the UI of the surveillance form to remove default integer values

JohnnieShen:
This PR introduced sanitization for numeric inputs in surveillance form. The inputs used to be taken as integers, so the default 0 would persist and inputs such as adding X to before or after the 0 will become X0 or 0X. Now, the initial 0 would work as a placeholder and any input would override it. There was also an issue with 2 inputs at the same time when the field is 0, where only one input would register. This issue have been resolved.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
@@ -159,12 +159,11 @@ fun SurveillanceFormScreen(
 
                 TextEntryField(
                     label = "Number of House Occupants",
-                    value = state.surveillanceForm.numPeopleSleptInHouse.toString(),
+                    value = if (state.surveillanceForm.numPeopleSleptInHouse == 0) "" else state.surveillanceForm.numPeopleSleptInHouse.toString(),
                     onValueChange = {
-                        onAction(SurveillanceFormAction.EnterNumPeopleSleptInHouse(
-                            sanitizeNumericInput(it, state.surveillanceForm.numPeopleSleptInHouse)
-                        ))
+                        onAction(SurveillanceFormAction.EnterNumPeopleSleptInHouse(it.filter { c -> c.isDigit() }))
                     },
+                    placeholder = "0",
                     singleLine = true,
                     keyboardType = KeyboardType.Number,
                 )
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vci/vectorcamapp/surveillance_form/presentation/SurveillanceFormScreen.kt
```diff
@@ -159,10 +159,11 @@ fun SurveillanceFormScreen(
 
                 TextEntryField(
                     label = "Number of House Occupants",
-                    value = state.surveillanceForm.numPeopleSleptInHouse.toString(),
+                    value = if (state.surveillanceForm.numPeopleSleptInHouse == 0) "" else state.surveillanceForm.numPeopleSleptInHouse.toString(),
                     onValueChange = {
-                        onAction(SurveillanceFormAction.EnterNumPeopleSleptInHouse(it))
+                        onAction(SurveillanceFormAction.EnterNumPeopleSleptInHouse(it.filter { c -> c.isDigit() }))
                     },
```
aryamanshodhan, line 165: What is c? Avoid using short forms and single character variable names?

aryamanshodhan (changes requested):
Well done Johnny. The code you have looks good. I have a question:
What happens if a user forgets to enter a number? It would be good to have validation checks in place for making sure the field is not blank just so that the user does'nt miss entering a datapoint. 
I like the idea of having a placeholder, but the only concern I would have is that the user might skip over the field thinking that it is already filled out when it is, in fact, not.

JohnnieShen:
> Well done Johnny. The code you have looks good. I have a question: What happens if a user forgets to enter a number? It would be good to have validation checks in place for making sure the field is not blank just so that the user does'nt miss entering a datapoint. I like the idea of having a placeholder, but the only concern I would have is that the user might skip over the field thinking that it is already filled out when it is, in fact, not.

Currently if no value have been filled out the value will be 0 if the placeholder is there and says 0. I am assuming that this will work since our default value was 0 anyways. Only different behavior is if the user deletes all input from a field, the old version would throw an error but the new version will continue as if the field is still 0. I couldn't come up with a better solution while keeping the screen stateless so this is as good as I can get. Please let me know if you have a better idea on how I could implement this!

aryamanshodhan (approved):

ThomasT0830 (approved):
Everything looks good to me! Tested on a physical device

PR #43, opened 2025-07-08
SOF-198: UI - Completed Session Details Screen

aryamanshodhan:
This PR introduces visual refinements to the Complete Session screen. Specimen tiles are now sized to fit 90% of the screen width, and when only one tile is present, it is centered to improve visual balance. If no specimens are available, an informative message is shown instead of leaving the space empty.

In addition, this update enables the display of session form details, giving users a clearer and more complete view of the session information alongside the specimens. Visual verification of the Ui was conducted for testing.


JohnnieShen (approved):
UI visually looks good! Only questions is whether we want to display something other than just blank (e.g. N/A) when an optional field is not filled out.

ThomasT0830 (approved):
The UI looks great!

PR #44, opened 2025-07-09
SOF-257: Implement bounding box UI as rendered on preview frame to remove conversion step

JohnnieShen:
This PR introduced a refactor that removed the BoundingBoxUI class. Now all bounding boxes will be drawn on a single canvas, which would improve performance, the boxes are also rendered relative to the preview frame instead of the whole screen, which makes moving the preview frame possible. Now obsolete methods such as ConvertToBoundingBox and ConvertToBoundingBoxUI are also removed and code are refactored respectively.

This PR should be tested extensively to make sure that after the refactor the app functions the same as before both visually and functionally.


app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -97,9 +95,9 @@ class ImagingViewModel @Inject constructor(
 
     private val _state = MutableStateFlow(ImagingState())
     val state: StateFlow<ImagingState> = combine(
-        specimensUiFlow, _state
-    ) { specimensUi, state ->
-        state.copy(capturedSpecimensAndBoundingBoxesUi = specimensUi)
+        specimensFlow, _state
+    ) { specimens, state ->
```
aryamanshodhan, line 99: lets rename specimensFlow to _specimensAndBoundingBoxes

app/src/main/java/com/vci/vectorcamapp/imaging/data/CameraManagerImplementation.kt
```diff
@@ -37,12 +37,17 @@ class CameraFocusManagerImplementation (
     }
 
 
-    override fun autoFocusOn(box: BoundingBoxUi) {
+    override fun autoFocusOn(box: BoundingBox) {
+        if (previewView.width == 0 || previewView.height == 0) {
+            Log.w("CameraFocusManager", "autoFocusOn(): previewView not ready yet")
+            return
```
aryamanshodhan: Is this the only way to check whether the preview view is ready or not?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
```diff
@@ -1,23 +1,38 @@
...
 import androidx.compose.ui.graphics.drawscope.Stroke
-import com.vci.vectorcamapp.imaging.presentation.model.BoundingBoxUi
+import androidx.compose.ui.unit.IntSize
+import com.vci.vectorcamapp.core.domain.model.BoundingBox
 
 @Composable
-fun BoundingBoxOverlay(boundingBoxUi: BoundingBoxUi, modifier: Modifier = Modifier) {
+fun BoundingBoxOverlay(
+    boundingBoxes: List<BoundingBox>,
+    overlaySize: IntSize,
```
aryamanshodhan, line 16: BoundingBoxOverlay should take in a single bounding box. From a semantics perspective, it makes more sense to iterate through the list of bounding boxes and display a BoundingBoxOverlay for each of them. Move the list of bounding boxes and iterate over them one layer above.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -93,12 +95,19 @@ fun LiveCameraPreviewPage(
             modifier = Modifier
                 .fillMaxWidth()
                 .aspectRatio(1f / aspectRatio)
+                .onSizeChanged { outerOverlaySize = it }
```
aryamanshodhan, line 98: Can you explain how this is working? How is the autofocus part related to displaying the size of the bounding box? I feel like I am understanding the code wrong.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
```diff
@@ -93,12 +95,19 @@ fun LiveCameraPreviewPage(
...
             AndroidView(
                 factory = { previewView },
                 modifier = Modifier.fillMaxSize()
             )
 
+            BoundingBoxOverlay(
+                boundingBoxes = boundingBoxes,
+                overlaySize = outerOverlaySize,
+                modifier = Modifier.fillMaxSize()
+            )
```
aryamanshodhan: Iterate through the list of bounding boxes here, and display each of them.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/specimen/gallery/CapturedSpecimenOverlay.kt
```diff
@@ -17,58 +18,82 @@ import androidx.compose.material3.Icon
...
+                        modifier = Modifier.fillMaxSize()
+                    )
+                }
 
-        BoundingBoxOverlay(
-            boundingBoxUi = boundingBoxUi, modifier = Modifier.fillMaxSize()
-        )
+                BoundingBoxOverlay(
+                    boundingBoxes = listOf(boundingBox),
+                    overlaySize = overlaySize,
```
aryamanshodhan: If you take a single bounding box as argument, you dont need to pass in listOf(boundingBox) which is a bit awkward given that you will only ever have one bounding box per specimen.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/specimen/gallery/CapturedSpecimenOverlay.kt
```diff
@@ -17,58 +18,82 @@ import androidx.compose.material3.Icon
...
+    boundingBox: BoundingBox,
     modifier: Modifier = Modifier,
     specimenBitmap: Bitmap? = null,
     onSpecimenIdCorrected: ((String) -> Unit)? = null,
     onRetakeImage: (() -> Unit)? = null,
     onSaveImageToSession: (() -> Unit)? = null,
 ) {
     val context = LocalContext.current
+    var overlaySize by remember { mutableStateOf(IntSize.Zero) }
+
```
aryamanshodhan, line 56: Again, can you explain how overlaySize works? You might benefit from BoxWithConstraints.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/specimen/gallery/CapturedSpecimenOverlay.kt
```diff
@@ -17,58 +18,82 @@ import androidx.compose.material3.Icon
...
-                contentDescription = specimen.id,
-                contentScale = ContentScale.Fit,
-                modifier = Modifier.fillMaxSize()
-            )
-        }
+        Box(
+            modifier = Modifier.fillMaxSize(),
+            contentAlignment = Alignment.Center
+        ) {
+            Box(
```
aryamanshodhan, line 67: Box within box within box lol. Is there a way to consolidate?

aryamanshodhan (changes requested):
Fantastic job Johnny! This is some very clean code! Can you please look through my comments? There are a few opportunities for consolidation and clean up. Thanks, this is truly well written! I will test on the device once you address these comments.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
JohnnieShen: My idea is that having one Canvas and drawing all boxes on it would have better performance, I could rename things to make things make more sense semantics wise but I stand by this choice since it is better than having multiple canvas.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreviewPage.kt
JohnnieShen: I am not sure if I am understanding the question correctly, but I don't think the two parts you mentioned are related, at least not explicitly related. In my current implementation I uses the OverlaySize to calculate both how the bounding boxes should look as well as where the camera should focus to hardware wise.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/specimen/gallery/CapturedSpecimenOverlay.kt
JohnnieShen: I could read into BoxWithConstraints but I wasn't familiar with it, I can see if switching to it would help me consolidate

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/specimen/gallery/CapturedSpecimenOverlay.kt
JohnnieShen: Unfortunately it might not be impossible to consolidate here. The first box holds the root level, second box centers the preview assembly and the third one I added becasue the manual focus ring won't appear at all without a internal box to group every (maybe due to some UI stacking issue)

aryamanshodhan (approved):
Looks good. As mentioned, I will likely optimize the code a but to remove the overlaySize state variable.

ThomasT0830 (approved):
Everything looks good to me!

PR #46, opened 2025-07-11
SOF-260: Display number of incomplete sessions in main screen

JohnnieShen:
This PR added the functionality of displaying the number of incomplete session as a badge on landing screen. The code makes sure that the badge won't overlap with other text even if the number is large. Code is tested on a physical device both with the actual imcomplete session count and arbitrary large numbers, in this case 1000.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingScreen.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingScreen.kt
@@ -63,7 +63,8 @@ fun LandingScreen(
                         title = "View Incomplete Sessions",
                         description = "Resume and complete any unfinished sessions.",
                         icon = painterResource(R.drawable.landing_incomplete_sessions_icon),
-                        onClick = { onAction(LandingAction.ViewIncompleteSessions) })
+                        onClick = { onAction(LandingAction.ViewIncompleteSessions) },
+                        badgeCount = state.incompleteSessionsCount)
 
                     LandingActionTile(
                         title = "View Complete Sessions",
--- a/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.viewModelScope
 import com.vci.vectorcamapp.core.domain.cache.CurrentSessionCache
 import com.vci.vectorcamapp.core.domain.cache.DeviceCache
 import com.vci.vectorcamapp.core.domain.repository.ProgramRepository
+import com.vci.vectorcamapp.core.domain.repository.SessionRepository
 import com.vci.vectorcamapp.core.presentation.CoreViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.channels.Channel
@@ -23,11 +24,13 @@ class LandingViewModel @Inject constructor(
     private val deviceCache: DeviceCache,
     private val currentSessionCache: CurrentSessionCache,
     private val programRepository: ProgramRepository,
+    private val sessionRepository: SessionRepository,
 ) : CoreViewModel() {
 
     private val _state = MutableStateFlow(LandingState())
     val state: StateFlow<LandingState> = _state.onStart {
         loadLandingDetails()
+        observeIncompleteSessionsCount()
     }.stateIn(
         viewModelScope, SharingStarted.WhileSubscribed(5000L), LandingState()
     )
@@ -94,6 +97,14 @@ class LandingViewModel @Inject constructor(
                 it.copy(
                     enrolledProgram = program, isLoading = false
                 )
+            }
+        }
+    }
+
+    private fun observeIncompleteSessionsCount() {
+        viewModelScope.launch {
+            sessionRepository.observeIncompleteSessions().collect { sessions ->
+                _state.update { it.copy(incompleteSessionsCount = sessions.size) }
             }
         }
     }
--- a/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
@@ -29,17 +29,18 @@ class LandingViewModel @Inject constructor(
     sessionRepository: SessionRepository,
 ) : CoreViewModel() {
 
-    private val _state = MutableStateFlow(LandingState())
-
     private val _incompleteSessionsCount = sessionRepository.observeIncompleteSessions()
         .map { it.size }
         .stateIn(viewModelScope, SharingStarted.WhileSubscribed(), 0)
 
+    private val _state = MutableStateFlow(LandingState())
     val state: StateFlow<LandingState> = combine(
         _state,
         _incompleteSessionsCount
     ) { state, incompleteSessionsCount ->
         state.copy(incompleteSessionsCount = incompleteSessionsCount)
+    }.onStart {
+        loadLandingDetails()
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
```diff
@@ -23,11 +24,13 @@ class LandingViewModel @Inject constructor(
...
     private val currentSessionCache: CurrentSessionCache,
     private val programRepository: ProgramRepository,
+    private val sessionRepository: SessionRepository,
 ) : CoreViewModel() {
 
     private val _state = MutableStateFlow(LandingState())
     val state: StateFlow<LandingState> = _state.onStart {
         loadLandingDetails()
+        observeIncompleteSessionsCount()
     }.stateIn(
```
aryamanshodhan, line 44: Do all the loading within one function, because thes functions all run asynchronously so this will introduce risk of race conditions.

app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingScreen.kt
```diff
@@ -63,7 +63,8 @@ fun LandingScreen(
                         title = "View Incomplete Sessions",
                         description = "Resume and complete any unfinished sessions.",
                         icon = painterResource(R.drawable.landing_incomplete_sessions_icon),
-                        onClick = { onAction(LandingAction.ViewIncompleteSessions) })
+                        onClick = { onAction(LandingAction.ViewIncompleteSessions) },
+                        badgeCount = state.incompleteSessionsCount)
```
aryamanshodhan, line 68: Figure out a way to display the badge without having to pass badgeCount as an argument, because not all LandingActionTiles should have an attribute called "badgeCount". There should be a way to display both together without having to pass it in as argument. I would suggest having a box with a Badge and Landing Action tile. Make the positioning of the badge absolute and at the top right corner of the incomplete sessions tile.

app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
```diff
@@ -97,4 +100,12 @@ class LandingViewModel @Inject constructor(
...
     }
+
+    private fun observeIncompleteSessionsCount() {
+        viewModelScope.launch {
+            sessionRepository.observeIncompleteSessions().collect { sessions ->
+                _state.update { it.copy(incompleteSessionsCount = sessions.size) }
+            }
+        }
+    }
 }
```
aryamanshodhan: Move this into the loadLandingDetails function. You should probably not collect the flow. Look into how we did it for the incomplete and complete sessions using the combine function.

app/src/main/java/com/vci/vectorcamapp/landing/presentation/components/LandingActionTile.kt
```diff
@@ -55,6 +60,23 @@ fun LandingActionTile(
...
+                            ),
+                        containerColor = MaterialTheme.colors.error
+                    ) {
+                        Text(
+                            text = badgeCount.toString(),
+                            color = MaterialTheme.colors.buttonText
+                        )
+                    }
+                }
             }
```
aryamanshodhan, line 79: Yeah badge should probably not belong here. You should have it in the landing screen. If in the future we add actions which require us to add more badges, we can consider moving it in here.

aryamanshodhan (changes requested):
Well done Johnny! Please review the changes here. Thanks!

app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingScreen.kt
JohnnieShen: I don't think there is a way around it if we want the same effect. A box with a Badge and Landing Action tile would result in the badge being on the top right of the entire landing action tile,  not the logo itself. Which imo would look worse and less intuitive. Because we have this level of abstraction for landing action tile we have to have this parameter in order to change how the inside looks.

aryamanshodhan: Thats fine. I was intending for it to be on the top right of the entire tile, kind of like unread notifications which appear at the top right of an app widget. Can you attach some screenshots of what it looks like?

app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
```diff
@@ -23,18 +26,33 @@ class LandingViewModel @Inject constructor(
...
 ) : CoreViewModel() {
 
     private val _state = MutableStateFlow(LandingState())
-    val state: StateFlow<LandingState> = _state.onStart {
-        loadLandingDetails()
+
+    private val _incompleteSessionsCount = sessionRepository.observeIncompleteSessions()
+        .map { it.size }
+        .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000L), 0)
+
```
aryamanshodhan, line 35: Dont need a 5000L timeout on this

app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
```diff
@@ -96,5 +114,12 @@ class LandingViewModel @Inject constructor(
                 )
             }
         }
+
+        viewModelScope.launch {
+            sessionRepository.observeIncompleteSessions().collect { sessions ->
+                _state.update { it.copy(incompleteSessionsCount = sessions.size) }
+            }
+        }
```
aryamanshodhan: Dont need this explicit call to observeIncompleteSessions. The call you make in the combine function will update the state.

app/src/main/java/com/vci/vectorcamapp/landing/presentation/components/LandingActionTile.kt
```diff
@@ -22,13 +25,15 @@ import com.vci.vectorcamapp.core.presentation.components.ui.ActionTile
 import com.vci.vectorcamapp.ui.extensions.colors
 import com.vci.vectorcamapp.ui.extensions.dimensions
 
+@OptIn(ExperimentalMaterial3Api::class)
```
aryamanshodhan: You dont need this OptIn

aryamanshodhan (changes requested):
Good job! Couple of opportunities for consolidation.

app/src/main/java/com/vci/vectorcamapp/landing/presentation/LandingViewModel.kt
```diff
@@ -23,18 +26,33 @@ class LandingViewModel @Inject constructor(
...
+    val state: StateFlow<LandingState> = combine(
+        _state,
+        _incompleteSessionsCount
+    ) { state, incompleteSessionsCount ->
+        state.copy(incompleteSessionsCount = incompleteSessionsCount)
     }.stateIn(
-        viewModelScope, SharingStarted.WhileSubscribed(5000L), LandingState()
+        viewModelScope,
+        SharingStarted.WhileSubscribed(),
+        LandingState()
```
aryamanshodhan, line 47: Sorry I wasnt clear in my previous comment. Typically the 5000L subscription time is required when defining only the state, not other private variables, so you need 5000L here. Here is a video explaining this:
https://youtu.be/mNKQ9dc1knI?si=f6FTP6pXyjwW4Gcl

aryamanshodhan (dismissed):
Well done Johnny! Approving the PR but please make the requested change before merging. Thanks!

JohnnieShen:
Thank you for the information! Learnt something today. Please re-review.

aryamanshodhan (approved):

ThomasT0830 (approved):

PR #48, opened 2025-07-13
SOF-245: Clarify concordance between app and PC models.

aryamanshodhan:
This pull request significantly refines the image preprocessing and model inference pipeline within the Android application. The main goal here is to boost how well our model results match between the Android app and PC-based models, directly tackling the bounding box coordinate differences we've seen. These variances popped up because of how JPEG compression loses data when images are saved, and also because of tiny floating-point precision differences when running inference on multi-threaded CPUs or GPUs.

This update introduces a unified and consistent end-to-end image processing and model post-processing pipeline. It covers standardized resizing, padding, normalization, NMS (Non-Maximum Suppression), and coordinate mapping, directly aligning app results with PC models. Model input and output tensor shapes are now dynamically retrieved from the loaded .tflite model, making the app more adaptable to future model changes. What's more, we've set up the TFLite interpreter to explicitly use all available CPU threads, giving us clear control over resource use and leading to more predictable performance across devices.

While this refactoring greatly improves how well our app's results match, it's good to know about a few specific situations that might still cause tiny differences:

Floating-Point Precision: You might still see minor pixel differences (around 1-2 pixels) in bounding box coordinates between Android and PC, especially if you're using a GPU delegate. This is just how floating-point math works in highly optimized deep learning setups on different hardware and software. 

Live Camera Feed vs. Saved Image Discrepancy: When you capture an image from the live camera and then save it (usually as a JPEG by default), some pixel data gets lost because JPEG is a lossy format. So, the original image from the camera isn't pixel-for-pixel identical to the one you load later. So, when you're testing, focus on how close the raw logit outputs are, rather than expecting exact matches, especially if you're comparing across different environments or storage states.

Bounding Box Coordinate Representation: Another reason for those 1-2 pixel differences between our VectorCamApp and the ModelTester app (from the provided repository) comes down to how we handle coordinates. Our VectorCamApp saves (x1, y1, w, h) as floating-point values, where x1 and y1 are the top-left coordinates. The ModelTester app, on the other hand, internally saves (x1, y1, x2, y2) as absolute integer values (top-left and bottom-right), and then calculates width and height from w = (x2 - x1) and h = (y2 - y1). This difference in using floats versus integers, and how numbers get rounded, can cause those small pixel discrepancies.

Let's look at an example based on a normalized YOLO output: x1 = 0.39224634, y1 = 0.5137039, w = 0.21548389, h = 0.20433636. For screen dimensions of 2448 by 3264:

VectorCamApp's (x1, y1, w, h) calculation:

x1 = (0.39224634 * 2448).toInt() = 960
y1 = (0.5137039 * 3264).toInt() = 1676
w = (0.21548389 * 2448).toInt() = 527
h = (0.20433636 * 3264).toInt() = 666

ModelTester app's (x1, y1, x2, y2) calculation:

x1 = (0.39224634 * 2448).toInt() = 960
y1 = (0.5137039 * 3264).toInt() = 1676
x2 = ((0.39224634 + 0.21548389) * 2448).toInt() = 1487
y2 = ((0.5137039 + 0.20433636) * 3264).toInt() = 2343
Calculated w = 1487 - 960 = 527
Calculated h = 2343 - 1676 = 667

You'll notice a 1-pixel difference in h (666 vs 667) due to how rounding happens at different points in each app's calculation. This difference is enough to cause COMPLETELY different logits for the ML inference.

Phone Models: MediaTek chipset phones act a little funky (A15, A32) because they will process differently based on their built in processor. However, you should still test for closeness of logits, and perhaps exactness when using the image picker on both codebases (see below).

To thoroughly check how well our model results line up, follow these steps:

Set up Your Ground Truth Source: Get the YOLOClassificationAPP repository ready from https://github.com/sailohitaksh-cryptic/YOLOClassificationAPP/tree/main. This PC app will be your standard for expected model output. Make sure both the Android app and the ModelTester app are using the exact same .tflite model file.

Configure the Android App for Testing:

Implement an Image Picker: Add a quick image picker to your ImagingScreen. This means creating an ImagingAction.PickImages that accepts an image URI, handling that action in your ImagingViewModel to get the Bitmap from the URI using loadBitmapFromUri, and adding a button and ActivityResultContracts.PickMultipleVisualMedia launcher to your ImagingScreen to start the picker. Happy to help you set this up - let me know.

Run the Full Pipeline: Once you get the Bitmap from the URI, make sure it goes through the entire inference pipeline: inferenceRepository.detectSpecimen(bitmap). This should mirror how ImagingAction.CaptureImage processes its Bitmap (including the in-memory JPEG degradation for consistency).

If you capture an image, remember to save it by adding it to the session so you can then use the image picker on the VectorCam app and the ModelTester app. You will notice that the outputs change when running inference immediately after capturing the image vs. loading from the file storage.

Compare and Verify Results:

Log Logit Output: For every test image, log the raw logit outputs from both the Android app (you can do this right before getDetectedBoxes) and the ModelTester app. Remember, you're looking for closeness, not perfect matches, especially if you're using GPU delegates or multiple CPU threads, because of floating-point behavior.

Exact Matches with Image Picker: When you use the image picker in the VectorCamApp to load an image that was previously captured and saved by the app, the SHA checksum of the Bitmap that gets passed to inferenceRepository.detectSpecimen should be identical to the SHA checksum of the Bitmap that was originally generated by createJpegDegradedBitmap during the capture flow. This confirms that your save/load pipeline is working consistently.

Addressing Bounding Box Pixel Differences: To specifically test and understand the 1-2 pixel differences caused by coordinate conversion:

Run a test image through the Vectorcam app's image picker pipeline and log the final (x1, y1, w, h) bounding box coordinates.

Then, manually calculate x2 = x1 + w and y2 = y1 + h to get (x1, y1, x2, y2) for the ModelTester app.

In the ModelTester app (from the repository), hardcode the bitmap crop coordinates on line 143 (val croppedBitmap = Bitmap.createBitmap(bitmap, x1, y1, x2 - x1, y2 - y1)) using the integer values you just calculated.

After hardcoding the crop, you should see identical logit outputs from both the Android app and the ModelTester app for the same input image. This will confirm that the model's core inference is producing concordant results, and any minor pixel differences are simply due to the differing coordinate truncation and representation methods.

PLEASE DO NOT PUSH ANY CHANGES YOU MAKE TO THE BRANCH WHILE TESTING, ESPECIALLY THE IMAGE PICKER!


JohnnieShen (approved):

ThomasT0830 (approved):
Overall, everything looks good to me! Occasionally, the bounding box coordinate is off by a pixel

PR #49, opened 2025-07-14
SOF-216: Fix discrepancy between Live camera inference and image picker.

aryamanshodhan:
This PR addresses a critical issue where the ML model was processing different image data than what was being saved to device storage, causing SHA-256 checksum mismatches between inference and saved files.
Root Cause: The application was compressing captured images to JPEG format twice - once during the CaptureImage action for ML processing, and again during the SaveImageToSession action for file storage. Even with identical quality settings 100%, JPEG compression is non-deterministic and can produce different byte sequences in different contexts, leading to different checksums for what should be identical images.
Solution: Implemented a single-compression approach where the bitmap is compressed to JPEG bytes only once during capture. These exact bytes are then used for both ML inference by decoding back to bitmap and file storage by writing directly to disk. This ensures that the ML model processes the exact same image data that gets saved to the camera roll.
Key Changes:
- Added currentImageBytes field to ImagingState to store the canonical JPEG representation
- Removed duplicate currentImage bitmap storage to eliminate memory redundancy
- Created saveImageBytes() method in CameraRepository to write bytes directly to storage
- Updated UI layer to decode bytes to bitmap on-demand using remember() for performance
- Modified CaptureImage action to compress once and use those bytes for all subsequent operations
This fix guarantees that inference results are based on the exact same image data that users see saved in their photo gallery, eliminating any potential discrepancies between processing and storage.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/imaging/data/repository/CameraRepositoryImplementation.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/imaging/data/repository/CameraRepositoryImplementation.kt
@@ -2,7 +2,6 @@ package com.vci.vectorcamapp.imaging.data.repository
 
 import android.content.ContentValues
 import android.content.Context
-import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Environment
 import android.provider.MediaStore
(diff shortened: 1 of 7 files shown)


JohnnieShen (dismissed):
LGTM

ThomasT0830 (approved):

PR #50, opened 2025-07-16
SOF-262: Upon entering the Imaging Screen, user should start at the camera preview page as opposed to the beginning of the gallery.

JohnnieShen:
This PR reorders the gallery and preview screen in imaging screen. Now the preview screen should be at the 0th index, so that the user don't have to scroll through all images to find it.


aryamanshodhan (changes requested):
I think all you need to do is change the initial page number in the pagerState. Right now it is set to 0.

JohnnieShen:
I believe setting initial page number here to 0 is correct. I made the change that all gallery pages will be to the right of camera preview screen, and preview should be at index 0. This is the simplist way to make the user land at preview screen after entering imaging screen without introducing too much unnecessary logic.

aryamanshodhan:
For the current logic, sure. But why not set the initial index to be the last page so the user always starts on the camera screen, and the images which were previously captured appear before the camera? Doesn't that achieve the same effect in a simpler manner with just a couples of lines changed?

JohnnieShen:
Sure, if that's what we want. I personally feel like implementation like this is more intuitive since the most important page have a fixed index, which would simplify the when logic. But if you want to implement this the other way I can do that. I think I was on the wrong path when I first started working on this, so I ended up with this logic

aryamanshodhan:
I'm in favor of the second approach where the camera is at the end because that is also what the current VectorCam app workflow is like. 

@ThomasT0830 Can you provide an opinion on this?

JohnnieShen:
That's a fair point. I can update it to match what we have in a bit.

aryamanshodhan (changes requested):
Great job Johnny! Just a quick change. You should scroll to the camera page so set the scrolling to size + 1. One small request to avoid merge conflicts with my branch. I am working on the UI of the imaging screen which is currently using the when statement instead of those nested if statements. Can you pls revert back to the when statement?

JohnnieShen:
Will do, I am actually working on adding the loading animation before the "scroll" finishes, should be done in a bit

JohnnieShen:
I've added a loading animation that only dissappears after the scroll and db collection have finished, to make it work (because to wait for the scroll to finish we have to render the HorizontalPager and loading animation at the same time, otherwise scrollToPage will never be finished) I had to wrap the pager and animation in a Box, could you take a look and see if this would work with the UI? if not I can get it changed

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -127,6 +133,10 @@ class ImagingViewModel @Inject constructor(
     fun onAction(action: ImagingAction) {
         viewModelScope.launch {
             when (action) {
+                is ImagingAction.InitializationComplete -> {
+                    _isUiReady.value = true
+                }
```
aryamanshodhan: Why not just update the isLoading state variable? I dont think you need a separate flag for isUiReady

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -47,54 +50,72 @@ fun ImagingScreen(
...
+                        manualFocusPoint = state.manualFocusPoint,
+                        onAction = onAction,
+                        modifier = Modifier,
+                        captureEnabled = !state.isProcessing
+                    )
+                }
             }
         }
+
+        if (state.isLoading) {
```
aryamanshodhan: I'd say move this to the navGraph. Because the LoadingAnimation is not a property of the imaging screen.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -47,54 +50,72 @@ fun ImagingScreen(
...
-            page < state.capturedSpecimensAndInferenceResults.size -> {
-                CapturedSpecimenOverlay(
-                    specimen = state.capturedSpecimensAndInferenceResults[page].specimen,
-                    inferenceResult = state.capturedSpecimensAndInferenceResults[page].inferenceResult,
-                    modifier = modifier
-                )
-            }
+    LaunchedEffect(state.capturedSpecimensAndInferenceResults.size) {
+        pagerState.scrollToPage(state.capturedSpecimensAndInferenceResults.size)
+        onAction(ImagingAction.InitializationComplete)
```
aryamanshodhan: InitializationComplete is not an appropriate name for this action. The initialization is happening only the first time this screen is loaded. However, this block of code runs every time state.capturedSpecimensAndInferenceResults.size changes. And every time it changes, it is not an initialization.

aryamanshodhan (dismissed):
Great job Johnny! This is written well. Just a couple of opportunities for consolidation. Thanks!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
JohnnieShen: Because we want the animation to end after both the database is collected and the initial scroll is finished, if we just have a isLoading flag and set it to false here we won't know if the db has finished loading or not. I don't think I ever figured out a way to do this with one flag only.

JohnnieShen:
Works on my device but I'm a bit unsure of how it works, is it still waiting for the scroll page animation to finish before taking down the animation?

ThomasT0830 (approved):

PR #53, opened 2025-07-18
SOF-265: Disable capture button when isCapturing 

ThomasT0830:
This pull request disables the `IconButton` in `LiveCameraPreviewPage` when `isCapturing` is `true`. A boolean parameter `captureEnabled`, which is set to `!state.isCapturing`, is passed from the `ImagingScreen` to the `LiveCameraPreviewPage`.

The code was only tested on the Samsung A15, so more testing on the Motorola is necessary.


aryamanshodhan (approved):
Good

aryamanshodhan (changes requested):
Make the button disappear when capturing

aryamanshodhan (approved):

PR #55, opened 2025-07-22
SOF-264: Clicking on capture image button right after surveillance form submission causes a fatal crash

JohnnieShen:
This PR added a guard at the beginning of CaptureImage that only becomes true after the first frame has been processed. This prevents a fatal crash caused by capturing an image while the camera is not ready.


aryamanshodhan (approved):
Great Job Johnny! Ready to merge in. A couple of optional recommendations I have:
1. Do not set isCameraReady to false after every image capture. Might be laggy if you need to wait for the first frame to process after saving the image, but this is minor.
2. Kotlin is weird when it comes to return statements because apparently it allows you to return at different locations within the function (Eg. with the return@launch). If I was in your place, instead of doing 

if (!_state.value.isCameraReady) return@launch

I would do:

if (_state.value.isCameraReady) { /* Body of if statement, and rest of the processing */}

ThomasT0830 (approved):
Everything looks good to me! Tested it on the Motorola

JohnnieShen:
Just so we can merge it in I'm not going to do another commit which would null the approval. Will return if issues arise.

PR #58, opened 2025-07-24
SOF-266: Incorrect start destination after granting, then enabling permissions

ThomasT0830:
This pull request fixes an issue where the app incorrectly navigates to the Registration screen after returning from the Settings app (e.g., when enabling GPS or permissions), even when a device was already registered. The issue was caused by a one-time DataStore read in `determineStartDestination()` of `MainViewModel`, which could return a default, uninitialized DTO when the app resumes, falsely indicating that the device is unregistered.

The fix replaces the one-shot initialization with a reactive Flow-based approach. A new `observeDevice()` method in `MainViewModel` listens to `DeviceCache.getDeviceFlow()`, updating `startDestination` only when the actual registration state changes. This eliminates timing issues and prevents overwriting the correct navigation state.

The `Device` to/from `DeviceDTO` mapping was also refactored to align with Clean Architecture. A new `DeviceMappers.kt` centralizes conversions using `toDomain()` and `toDto()`. `programId == -1` is now the source of truth for determining whether a `Device` is registered. `DeviceCacheImplementation` was updated to use these mappers and provide `getDeviceFlow()` and `getProgramIdFlow()` for reactive observation.

The code has been thoroughly tested on a physical device.


app/src/main/java/com/vci/vectorcamapp/core/data/mappers/DeviceMapper.kt
```diff
@@ -0,0 +1,27 @@
...
+
+fun Device.toDto(programId: Int): DeviceCacheDto {
+    return DeviceCacheDto(
+        id = id,
+        programId = programId,
+        model = model,
+        registeredAt = registeredAt,
+        submittedAt = submittedAt
+    )
+}
```
aryamanshodhan: I like the thought process, but the Device.toDto function is not clear. There are 4 different types of DTOs for the device, DeviceCacheDto, DeviceDto, PostDeviceRequestDto, PostDeviceResponseDto. By saying toDto whenever you use this function, I would not know which Dto you are trying to convert to. I would think you are converting the Device to DeviceDto. I think it is more intuitive to just create the DeviceCacheDto object within the cacheImplementation

app/src/main/java/com/vci/vectorcamapp/core/domain/cache/DeviceCache.kt
```diff
@@ -7,4 +7,7 @@ interface DeviceCache {
     suspend fun saveDevice(device: Device, programId: Int)
     suspend fun getDevice(): Device?
     suspend fun getProgramId(): Int?
+
+    fun getDeviceFlow(): Flow<Device?>
+    fun getProgramIdFlow(): Flow<Int?>
```
aryamanshodhan: Instead of getDeviceFlow and getProgramIdFlow, call these functions observeDevice, and observeProgramId

app/src/main/java/com/vci/vectorcamapp/main/presentation/MainViewModel.kt
```diff
@@ -22,13 +23,15 @@ class MainViewModel @Inject constructor(
 ) : CoreViewModel() {
 
     private val _state = MutableStateFlow(MainState())
-    val state = _state.onStart {
-        determineStartDestination()
-    }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), MainState())
+    val state = _state.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), MainState())
```
aryamanshodhan: _state.onStart{observeDevice}.stateIn(...)

app/src/main/java/com/vci/vectorcamapp/main/presentation/MainViewModel.kt
```diff
@@ -22,13 +23,15 @@ class MainViewModel @Inject constructor(
...
 
     private val _events = Channel<MainEvent>()
     val events = _events.receiveAsFlow()
 
+    init {
+        observeDevice()
+    }
+
     fun onAction(action: MainAction) {
```
aryamanshodhan: Move this out of the init block

app/src/main/java/com/vci/vectorcamapp/main/presentation/MainViewModel.kt
```diff
@@ -64,18 +67,21 @@ class MainViewModel @Inject constructor(
         }
     }
 
-    private fun determineStartDestination() {
+    private fun observeDevice() {
```
aryamanshodhan: observeDevice here would shadow the cache function deviceCache.observeDevice. Rename to something else.

app/src/main/java/com/vci/vectorcamapp/main/presentation/MainViewModel.kt
```diff
@@ -64,18 +67,21 @@ class MainViewModel @Inject constructor(
...
-                emitError(MainError.DEVICE_FETCH_FAILED)
-            }
+                .onEach { device ->
+                    if (device != null && _state.value.startDestination != Destination.Landing) {
+                        _state.update { it.copy(startDestination = Destination.Landing) }
+                    } else if (device == null && _state.value.startDestination != Destination.Registration) {
+                        _state.update { it.copy(startDestination = Destination.Registration) }
+                    }
+                }
+                .collect { }
```
aryamanshodhan, line 83: I wonder if you can do something along the lines of how we load the incomplete and complete sessions, and then combine it with the _state using the combine function. See if something like that is possible. That might help get rid of this observeDevice function altogether.

aryamanshodhan (changes requested):
Well done Thomas, there are a few changes I would like you to implement. Please do ahead and make the changes. Thanks!

app/src/main/java/com/vci/vectorcamapp/core/data/cache/DeviceCacheImplementation.kt
```diff
@@ -23,25 +26,45 @@ class DeviceCacheImplementation @Inject constructor(
     }
 
     override suspend fun getDevice(): Device? {
-        val deviceCacheDto = dataStore.data.firstOrNull()
-        return if (deviceCacheDto == null || deviceCacheDto == DeviceCacheDto()) {
+        val dto = dataStore.data.first()
```
aryamanshodhan: This will throw an error if device is null...Unfortunately I think you will need a check for whether the dto is null

app/src/main/java/com/vci/vectorcamapp/core/data/cache/DeviceCacheImplementation.kt
```diff
@@ -23,25 +26,45 @@ class DeviceCacheImplementation @Inject constructor(
...
-                submittedAt = deviceCacheDto.submittedAt
+                id = dto.id,
+                model = dto.model,
+                registeredAt = dto.registeredAt,
+                submittedAt = dto.submittedAt
             )
         }
     }
 
     override suspend fun getProgramId(): Int? {
```
aryamanshodhan: Can you explain why you decided to migrate to using the program ID instead of checking to see if the device cache dto is null?

aryamanshodhan (changes requested):

aryamanshodhan (approved):

JohnnieShen (approved):

PR #64, opened 2025-07-27
SOF-242: Data Collection Mode

aryamanshodhan:
This PR introduces support for Data Collection Mode in VectorCam, enabling sessions where the surveillance form is not required. Unlike Surveillance Mode, which mandates form input, Data Collection Mode bypasses form collection entirely — based not just on session type but on broader program-level logic. This is especially useful for RCT arms or special deployments. To maintain modularity and flexibility, we extended the Strategy + Factory pattern already used in the imaging layer. A new SurveillanceFormWorkflow abstraction now governs form behavior, and the IntakeViewModel uses the factory to determine whether a form should be displayed, removing the need for additional booleans in UI state.

In parallel, we refined the ImagingWorkflow architecture to support Data Collection Mode more cleanly. Previously, all workflows assumed form capture as a prerequisite. With this update, we added new ImagingWorkflow variants that account for the absence of a form — for example, skipping form validation steps, image tagging logic, or metadata handling that previously depended on form inputs. This ensures consistent separation of session-specific behavior while maintaining the existing ViewModel contract.

We also made several changes to the local Room database schema. SpecimenEntity now includes a nullable remoteId field to track server-assigned IDs, and both SpecimenImageEntity and InferenceResultEntity now use String IDs instead of UUID, simplifying backend sync and improving consistency. Related foreign keys and indices were updated, and new type converters for UUID and FloatArray were added to support this transition. The migration (MIGRATION_9_10_REFACTOR_SPECIMEN_IMAGE_AND_INFERENCE_RESULT_KEYS) is destructive but safe given the current user base. On the networking side, we’ve configured kotlinx.serialization to use explicitNulls = false so that inference results are omitted entirely from the payload when not present, rather than sending "inferenceResult": null" — ensuring backend compatibility and cleaner requests.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -36,7 +36,7 @@ android {
 
     buildTypes {
         debug {
-            buildConfigField("String", "BASE_URL", "\"https://api.vectorcam.org/\"")
+            buildConfigField("String", "BASE_URL", "\"https://test.api.vectorcam.org/\"")
         }
 
         release {
--- a/app/schemas/com.vci.vectorcamapp.core.data.room.VectorCamDatabase/10.json
+++ b/app/schemas/com.vci.vectorcamapp.core.data.room.VectorCamDatabase/10.json
@@ -1,0 +1,593 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 10,
+    "identityHash": "1f08119693f5da9dcd758c65cd0801ad",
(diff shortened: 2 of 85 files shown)


JohnnieShen (dismissed):

JohnnieShen (approved):

ThomasT0830 (approved):
Everything looks good to me!

PR #82, opened 2025-08-12
SOF-299: Specify build version name in the splash screen

aryamanshodhan:
This PR includes a very simple addition of the build version to the splash screen so that the user is able to see which version number they installed.


JohnnieShen (approved):
LGTM!

PR #84, opened 2025-08-15
SOF-300: Make notification IDs for data upload unique for each session

aryamanshodhan:
This PR solves the issue of notification updates for multiple sessions occurring on the same notification box. With this PR, each session gets its own notification displayed in a way that they are independent from other notification IDs. This involves using the sessionId's hashcode as the unique ID for the notification.


JohnnieShen (approved):

ThomasT0830 (approved):
Everything looks good to me!

PR #85, opened 2025-08-16
SOF-175: Optimize camera preview to reduce the amount of stutter

JohnnieShen:
This PR replaced the Tensorflow lite library with the more modern LiteRT library, it also addresses issues regarding preview UI stutter by moving tasks off the UI thread. Additional optimizations have also been implemented regarding the ProcessFrame function.


gradle/libs.versions.toml
```diff
@@ -27,6 +27,7 @@ room = "2.7.1"
 sentry = "5.7.0"
 tensorflowLite = "2.16.1"
 tensorflowLiteSupport = "0.4.4"
```
aryamanshodhan: Remove unused tflite dependencies

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/SpecimenImageAnalyzer.kt
```diff
@@ -1,13 +1,28 @@
...
+        const val FRAME_SKIP_RATE = 15
+    }
+
+    private var frameCounter = 0
 
     override fun analyze(frame: ImageProxy) {
-        onFrameReady(frame)
+        try {
+            if (frameCounter % FRAME_SKIP_RATE == 0) {
+                val bmp = frame.toUprightBitmap()
```
aryamanshodhan: bitmap, not bmp

app/src/main/java/com/vci/vectorcamapp/imaging/domain/strategy/concrete/SurveillanceImagingWorkflow.kt
```diff
@@ -110,5 +111,6 @@ class SurveillanceImagingWorkflow @Inject constructor(
 
     override fun close() {
         inferenceRepository.closeResources()
+        GpuDelegateManager.close()
```
aryamanshodhan: You do not need this. inferenceRepository.closeResources() already closes the GPU

app/src/main/java/com/vci/vectorcamapp/imaging/data/TfLiteSpecimenDetector.kt
```diff
@@ -273,6 +288,8 @@ class TfLiteSpecimenDetector(
                 try {
                     detector?.close()
                     detector = null
+                    gpuDelegate?.close()
+                    gpuDelegate = null
```
aryamanshodhan: Again, you dont need this since the delegate manager already closes the resources

app/src/main/java/com/vci/vectorcamapp/imaging/data/TfLiteSpecimenDetector.kt
```diff
@@ -264,6 +268,17 @@ class TfLiteSpecimenDetector(
         return finalInferenceResults
     }
 
+    private fun warmDetector() {
+        if (gpuDelegate != null) {
+            val inputShape = detector!!.getInputTensor(0).shape()
```
aryamanshodhan: detector!!.getInputTensor(0).shape()

Do not use !!
If for some reason the detector is null, this will crash the app. Replace with 
detector?.getInputTensor(0).shape()

Same for detector!!.getOutputTensor(0).shape()

aryamanshodhan: Also this code never runs because gpuDelegate is always null. You never reassign it in this file, so it is always null.

app/src/main/java/com/vci/vectorcamapp/imaging/data/TfLiteSpecimenDetector.kt
```diff
@@ -34,6 +36,7 @@ class TfLiteSpecimenDetector(
 ) : SpecimenDetector {
 
     private var detector: Interpreter? = null
+    private var gpuDelegate: GpuDelegate? = null
```
aryamanshodhan: This is never truly used in this file

app/src/main/java/com/vci/vectorcamapp/imaging/data/repository/CameraRepositoryImplementation.kt
```diff
@@ -29,19 +29,22 @@ class CameraRepositoryImplementation @Inject constructor(
...
 ) : CameraRepository {
     override suspend fun captureImage(controller: LifecycleCameraController): Result<ImageProxy, ImagingError> {
-        return suspendCoroutine { continuation ->
-            controller.takePicture(
-                ContextCompat.getMainExecutor(context), object : OnImageCapturedCallback() {
-                    override fun onCaptureSuccess(image: ImageProxy) {
-                        super.onCaptureSuccess(image)
-                        continuation.resume(Result.Success(image))
-                    }
+        return withContext(Dispatchers.Main) {
```
aryamanshodhan: Why do you need withContext? I think by default, this happens on the main branch

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -137,33 +139,30 @@ class ImagingViewModel @Inject constructor(
                 }
 
                 is ImagingAction.ProcessFrame -> {
-                    if (!_state.value.isCameraReady) {
-                        _state.update { it.copy(isCameraReady = true) }
-                    }
+                    viewModelScope.launch(Dispatchers.Default) {
```
aryamanshodhan: Is this needed? Does it cause a difference? Don't we already switch the context in the detect function?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -194,45 +193,45 @@ class ImagingViewModel @Inject constructor(
...
-                                        species = result.species,
-                                        sex = result.sex,
-                                        abdomenStatus = result.abdomenStatus
-                                    ),
-                                    currentImageBytes = jpegByteArray,
-                                    currentInferenceResult = result.capturedInferenceResult,
-                                    previewInferenceResults = emptyList()
-                                )
+                    viewModelScope.launch(Dispatchers.Default) {
+                        _state.update { it.copy(isProcessing = true) }
```
aryamanshodhan: State updates should happen on the main thread I believe

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -194,45 +193,45 @@ class ImagingViewModel @Inject constructor(
...
+                                emitError(error, SnackbarDuration.Short)
                             }
                         }.onError { error ->
-                            emitError(error, SnackbarDuration.Short)
-                        }
-                    }.onError { error ->
-                        if (error == ImagingError.NO_ACTIVE_SESSION) {
-                            _events.send(ImagingEvent.NavigateBackToLandingScreen)
+                            if (error == ImagingError.NO_ACTIVE_SESSION) {
+                                viewModelScope.launch { _events.send(ImagingEvent.NavigateBackToLandingScreen) }
```
aryamanshodhan: Where did this viewModelScope.launch come from? You are launching scopes from within scopes which is not good!

aryamanshodhan (changes requested):
Hey Johnny! Well done with this! This is written very cleanly, changes seem to be relatively straightforward! There are just a couple of concerns from my end:
1. Make sure the use of threading is consistent and follow best practices. I am not sure yet. Check out my comments. 
2. GPU delegate usage is incorrect and seems unused. Make sure to be EXTREMELY careful about this error: OpenGL ES API with no current context
If you see this error, in the logs, your GPU usage is NOT correct.
3. Make sure the model performance is identical to the PC. Logits from this app version should match the modeltester app EXACTLY. 
Moving forward, we should make this part of our regular testing pipeline. Make sure that the app matches the PC, especially when you change any files related to the imaging screen!

aryamanshodhan (approved):

PR #86, opened 2025-08-17
SOF-301: Add animation before image capture

aryamanshodhan:
This PR involves the creation of a capture animation which gets displayed every time the capture button is clicked on the imaging screen. It provides clear feedback to the user to hold the phone still while capture is happening.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/ui/theme/Colors.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/ui/theme/Colors.kt
@@ -32,6 +32,9 @@ data class Colors(
     val cardBackground: Color = Color(0xFFFFFFFF),
     val appBackground: Color = Color(0xFFF7F8F7),
     val transparent: Color = Color(0x00000000),
+    val animationScanColor: Color = Color(0xFF00FF7F),
+    val animationOverlayColor: Color = Color(0x80000000),
+    val animationCornerColor: Color = Color(0xFFFFFFFF)
 )
 
 val LocalColors = staticCompositionLocalOf { Colors() }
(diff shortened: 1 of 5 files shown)


ThomasT0830 (dismissed):
Looks good to me! Tested it on a Motorola

app/src/main/java/com/vci/vectorcamapp/ui/theme/Colors.kt
```diff
@@ -32,6 +32,9 @@ data class Colors(
     val cardBackground: Color = Color(0xFFFFFFFF),
     val appBackground: Color = Color(0xFFF7F8F7),
     val transparent: Color = Color(0x00000000),
+    val animationScanColor: Color = Color(0xFF00FF7F),
+    val animationOverlayColor: Color = Color(0x80000000),
+    val animationCornerColor: Color = Color(0xFFFFFFFF)
 )
```
aryamanshodhan, line 36: The purpose of the colors theme is to have reusable colors. I would recommend not having colors for individual elements because then our list of colors will keep growing indefinitely. I would recommend using one of the colors already available in the pallete, or defining a generic color name which you think will be used for multiple components in the future.

app/src/main/java/com/vci/vectorcamapp/ui/theme/Dimensions.kt
aryamanshodhan, line 1: The purpose of the dimensions in the theme is to have reusable dimensions. I would recommend not having dimensions for individual elements because then our list of dimensions will keep growing indefinitely. I would recommend using one of the dimensions already available, or defining a generic dimension name which you think will be used for multiple components in the future.

aryamanshodhan (commented):
Hey Johnny! Can you please look through the comments in these files and address them? It needs some minor cleanup. Thanks!

aryamanshodhan (commented):
Looks good to me

aryamanshodhan:
@ThomasT0830 You will need to approve this PR

JohnnieShen (approved):

ThomasT0830 (approved):

PR #89, opened 2025-08-21
SOF-268: Delete incomplete sessions

ThomasT0830:
This pull request allows users to permanently delete an incomplete session along with all associated images from device storage. To support this functionality, a new method `getImageUrisBySessionId()` has been added to the `SessionRepository`. This method retrieves all image URIs linked to a given session through a newly introduced DAO query.

The `IncompleteSessionViewModel` has been extended to implement the full deletion workflow. This workflow involves retrieving the session's image URIs, attempting to delete each image from device storage using the existing `CameraRepository`, and finally removing the session record from the database. 

On the UI side, the `IncompleteSessionScreen` has been updated to display a confirmation dialog before deletion. This dialog aligns with the app's existing design conventions and provides clear warnings to the user. Additionally, the `IncompleteSessionCard` and `ActionTile` components have been modified to support swipe-to-reveal behavior. Users can now swipe left on a session card in the `IncompleteSessionScreen` to reveal the delete button, which, when tapped, triggers the confirmation dialog. This interaction enhances usability while keeping destructive actions intentionally hidden until explicitly requested.

Testing was conducted on a Motorola device to ensure that the deletion workflow functions as expected, including verifying that the images associated with a deleted incomplete session are removed from Google Photos.


JohnnieShen (dismissed):
LGTM

app/src/main/java/com/vci/vectorcamapp/core/data/repository/SessionRepositoryImplementation.kt
```diff
@@ -60,6 +62,12 @@ class SessionRepositoryImplementation @Inject constructor(
         }
     }
 
+    override suspend fun getImageUrisBySessionId(sessionId: UUID): List<Uri> {
+        return sessionDao.getImageUrisBySessionId(sessionId).map { uriString ->
+                uriString.toUri()
```
aryamanshodhan: You do not need this mapping function. We have already defined a type converter for Room that will automatically convert URIs to strings when storing, and back to URIs when retrieving.

app/src/main/java/com/vci/vectorcamapp/core/presentation/components/tile/ActionTile.kt
aryamanshodhan, line 1: Why do you need to define onDelete in this file? deletion is specific to the incompleteSessionCard. I think any modifier properties related to enabling sliding should be passing in using this card's modifier property. If you notice, we use a lowercase "modifier" for this file to define the card's modifier. Therefore, any modifier properties you apply outside of this file whenever this component is rendered will pass onto this card. Therefore I think this file should remain unchanged, and onDelete should be specific to the incomplete session card.

app/src/main/java/com/vci/vectorcamapp/incomplete_session/presentation/components/IncompleteSessionCard.kt
```diff
@@ -40,6 +42,7 @@ fun IncompleteSessionCard(
 
     ActionTile(
         onClick = onClick,
+        onDelete = onDelete,
```
aryamanshodhan: In my opinion, ActionTile should not need a delete functionality.

app/src/main/java/com/vci/vectorcamapp/incomplete_session/presentation/IncompleteSessionScreen.kt
```diff
@@ -23,8 +30,58 @@ fun IncompleteSessionScreen(
...
+                Text(
+                    text = "This will permanently delete the session and all associated images from your device. This action cannot be undone.",
+                    style = MaterialTheme.typography.bodyMedium,
+                    color = MaterialTheme.colors.textSecondary
+                )
+            },
+            confirmButton = {
+                Button(
+                    onClick = {
+                        onAction(IncompleteSessionAction.ConfirmDeleteSession(state.deleteDialogSessionId!!))
```
aryamanshodhan: Do you need the !! here? You already have a check above this to check if the ID is null. !! is considered bad practice.

aryamanshodhan (changes requested):
Cleanly written Thomas! Can you pls address the comments and submit another review request? Thanks!

aryamanshodhan:
Please resolve the conflicts as well!

app/src/main/java/com/vci/vectorcamapp/incomplete_session/presentation/components/IncompleteSessionCard.kt
```diff
@@ -26,61 +40,115 @@ import com.vci.vectorcamapp.ui.extensions.colors
...
     onClick: () -> Unit,
+    onDelete: () -> Unit,
     modifier: Modifier = Modifier
 ) {
+    var offsetX by remember { mutableFloatStateOf(0f) }
+    val density = LocalDensity.current
+    val deleteWidthPx = with(density) { (MaterialTheme.dimensions.spacingExtraLarge * 2).toPx() }
+    val deleteWidthDp = with(density) { deleteWidthPx.toDp() }
+    val maxOffset = -deleteWidthPx
+    val shape = RoundedCornerShape(MaterialTheme.dimensions.cornerRadiusMedium)
```
aryamanshodhan: Maybe move the shape to the individual modifier fields of the components needing the shape. This variable feels a little misplaced.

aryamanshodhan (changes requested):
Great job Thomas! I think the code looks good. Please address the merge conflict and re-request review. Thanks!

app/src/main/java/com/vci/vectorcamapp/incomplete_session/presentation/components/IncompleteSessionCard.kt
```diff
@@ -31,56 +32,64 @@ import java.util.Locale
...
     onClick: () -> Unit,
+    onDelete: () -> Unit,
     modifier: Modifier = Modifier
 ) {
-
     val titleFormatter = remember { SimpleDateFormat("MMM dd, yyyy", Locale.getDefault()) }
-    val detailFormatter =
-        remember { SimpleDateFormat("MMM dd, yyyy 'at' h:mm a", Locale.getDefault()) }
+    val detailFormatter = remember { SimpleDateFormat("MMM dd, yyyy 'at' h:mm a", Locale.getDefault()) }
+    val deleteWidth = MaterialTheme.dimensions.spacingExtraLarge * 2
```
aryamanshodhan: Maybe convert this to a dimension in MaterialTheme Dimensions?

aryamanshodhan (changes requested):
Nice job

aryamanshodhan (approved):
Well done Thomas! I tested this on the A32 with sessions containing no image, 1 image, and >1 image. Seems to work.

JohnnieShen (approved):

PR #90, opened 2025-08-21
SOF-288: Fix camera initialization issue

ThomasT0830:
This pull request resolves crashes on Sentry with the error message "IllegalStateException: Not in application's main thread" that occurred when camera operations were called from background threads. `CameraX` requires all camera operations to run on the main thread, but the current code calls `cameraRepository.captureImage()` directly in the coroutine scope without guaranteeing that it runs in the main thread. The fix wraps all camera operations in `withContext(Dispatchers.Main)` to ensure they execute on the main thread, then switches to background threads for image processing, and returns to the main thread for UI updates. This pattern has been applied to both `CaptureImage` and `ProcessFrame` actions.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
@@ -139,34 +139,31 @@ class ImagingViewModel @Inject constructor(
                 }
 
                 is ImagingAction.ProcessFrame -> {
-                    withContext(Dispatchers.Main) {
-                        try {
-                            if (!_state.value.isCameraReady) {
-                                _state.update { it.copy(isCameraReady = true) }
-                            }
-                            withContext(Dispatchers.Default) {
-                                val bitmap = action.frame.toUprightBitmap()
-                                val liveFrameProcessingResult = imagingWorkflow.processLiveFrame(bitmap)
-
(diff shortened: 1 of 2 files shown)


JohnnieShen (dismissed):

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -139,31 +139,34 @@ class ImagingViewModel @Inject constructor(
...
-
-                        val liveFrameProcessingResult =
-                            imagingWorkflow.processLiveFrame(bitmap)
-
-                        validateSpecimenIdUseCase(
-                            liveFrameProcessingResult.specimenId, shouldAutoCorrect = true
-                        ).onSuccess { correctedSpecimenId ->
-                            _state.update {
-                                it.copy(currentSpecimen = it.currentSpecimen.copy(id = correctedSpecimenId))
+                    withContext(Dispatchers.Main) {
```
aryamanshodhan: We are already in the main branch here though, because we have viewModelScope.launch at the beginning of this function which launches a coroutine on the main branch.

aryamanshodhan (changes requested):
Hey Thomas! Are you sure about this fix? I think the code is already on the main thread.

vci-developers (approved):

aryamanshodhan (approved):

PR #94, opened 2025-08-22
SOF-305: Search and filter specimens within an incomplete/complete session

JohnnieShen:
This PR implements a search text bar inside the specimen page in completed session details page. The user is able to input queries and respective results will be displayed. In the query, spaces will be treated as AND operators and commas will be treated as OR opeartors, so if the user inputs "A, B C", it will be parsed to search for A OR B AND C.


app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/util/SpecimenSearchUtils.kt
aryamanshodhan, line 1: Can you create a more generic searchUtils class so that all subsequent search and filter tasks can be performed using this helper? Avoid having files with multiple standalone functions. I would just make it part of an object or a class.

Make it so that I can also use something similar for the session. Make it take a string as input

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/components/specimens/CompleteSessionSpecimens.kt
aryamanshodhan, line 1: This file needs to be significantly cleaned up. All searching and querying has to go into the view model. Use state for the entered query string. All functions for search need to come from the view model. No business logic in the components. The components are supposed to be dumb and only display/render stuff on the screen.

aryamanshodhan (changes requested):
Well done Johnny! Still needs some cleanup. Check out my comments.

ThomasT0830 (commented):
Overall, it looks great! Currently, however, when I perform a search and no result appears (no specimen matches the search), and I clear the search bar, it seems like no specimen reappears. The screen continues to read "No matching specimens found." Also, maybe we can update the UI of the search bar to match the rest of the screens in a different task

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsAction.kt
```diff
@@ -5,4 +5,6 @@ import com.vci.vectorcamapp.complete_session.details.presentation.enums.Complete
 sealed interface CompleteSessionDetailsAction {
     data object ReturnToCompleteSessionListScreen : CompleteSessionDetailsAction
     data class ChangeSelectedTab(val selectedTab: CompleteSessionDetailsTab) : CompleteSessionDetailsAction
+    data class UpdateQuery(val value: String) : CompleteSessionDetailsAction
+    data object ExecuteQuery : CompleteSessionDetailsAction
```
aryamanshodhan: UpdateSearchQuery(val query: String)
ExecuteSearch/PerformSearch

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsScreen.kt
```diff
@@ -62,8 +62,8 @@ fun CompleteSessionDetailsScreen(
                 )
 
                 CompleteSessionDetailsTab.SESSION_SPECIMENS -> CompleteSessionSpecimens(
-                    session = state.session,
-                    specimensWithImagesAndInferenceResults = state.specimensWithImagesAndInferenceResults,
+                    state = state,
+                    onAction = onAction,
```
aryamanshodhan: You should not pass state and onAction into any component other than the main screen

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsState.kt
```diff
@@ -39,4 +40,7 @@ data class CompleteSessionDetailsState(
     ),
     val surveillanceForm: SurveillanceForm? = null,
     val specimensWithImagesAndInferenceResults: List<SpecimenWithSpecimenImagesAndInferenceResults> = emptyList(),
+    val searchQuery: String = "",
+    val executedQuery: String = "",
+    val filteredSpecimenImageItems: List<SpecimenImageItem> = emptyList()
```
aryamanshodhan: You should not have a separate list for filtered images. You should just filter the original list. 

Also what is the difference between a searchQuery and and executedQuery? Why do you have state for that?

app/src/main/java/com/vci/vectorcamapp/core/presentation/util/search/SearchUtils.kt
```diff
@@ -0,0 +1,38 @@
+package com.vci.vectorcamapp.core.presentation.util.search
+
+import java.text.Normalizer
+
+object SearchUtils {
+    private fun normalize(text: String): String {
+        val lower = text.lowercase().trim()
+        val decomp = Normalizer.normalize(lower, Normalizer.Form.NFD)
```
aryamanshodhan: What is decomp?

app/src/main/java/com/vci/vectorcamapp/complete_session/domain/model/SpecimenImageItem.kt
```diff
@@ -0,0 +1,10 @@
+package com.vci.vectorcamapp.complete_session.domain.model
+
+import com.vci.vectorcamapp.core.domain.model.Specimen
+import com.vci.vectorcamapp.core.domain.model.SpecimenImage
+
+data class SpecimenImageItem(
+    val specimen: Specimen,
+    val specimenImage: SpecimenImage,
+    val badgeText: String
+)
```
aryamanshodhan: Explain this data class

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
```diff
@@ -44,15 +45,44 @@ class CompleteSessionDetailsViewModel @Inject constructor(
...
-        _specimensWithImagesAndInferenceResults, _state
-    ) { specimensWithImagesAndInferenceResults, state ->
-        state.copy(
-            specimensWithImagesAndInferenceResults = specimensWithImagesAndInferenceResults
-        )
-    }.onStart {
-        loadCompleteSessionDetails()
-    }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000L), CompleteSessionDetailsState())
+    val state: StateFlow<CompleteSessionDetailsState> =
+        combine(_specimensWithImagesAndInferenceResults, _state) { specimensWithImagesAndInferenceResults, current ->
```
aryamanshodhan: rename current to state

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
```diff
@@ -44,15 +45,44 @@ class CompleteSessionDetailsViewModel @Inject constructor(
...
+                            item.specimenImage.sex,
+                            item.specimenImage.abdomenStatus
+                        )
+                    )
+                }
+            }
+
+            current.copy(
+                specimensWithImagesAndInferenceResults = specimensWithImagesAndInferenceResults,
+                filteredSpecimenImageItems = filtered
```
aryamanshodhan: Maintaining 2 states here seems redundant

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
```diff
@@ -67,6 +97,12 @@ class CompleteSessionDetailsViewModel @Inject constructor(
                 is CompleteSessionDetailsAction.ChangeSelectedTab -> {
                     _state.update { it.copy(selectedTab = action.selectedTab) }
                 }
+                is CompleteSessionDetailsAction.UpdateQuery -> {
```
aryamanshodhan: UpdateSearchQuery might be a better name here

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/components/specimens/CompleteSessionSpecimens.kt
```diff
@@ -1,54 +1,93 @@
...
-    } else {
-        LazyRow(
-            modifier = modifier.fillMaxSize(), horizontalArrangement = Arrangement.Center
-        ) {
-            items(items = specimensWithImagesAndInferenceResults.asReversed()) { specimenWithSpecimenImagesAndInferenceResults ->
-                val specimen = specimenWithSpecimenImagesAndInferenceResults.specimen
-                val imageList =
-                    specimenWithSpecimenImagesAndInferenceResults.specimenImagesAndInferenceResults
-                val totalImages = imageList.size
+        return
```
aryamanshodhan: I dont like having return statements inside composables. Replace with a when statement or bring back the else statement

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/components/specimens/CompleteSessionSpecimens.kt
```diff
@@ -1,54 +1,93 @@
...
+            label = { Text("Search by specimen ID, species, etc.") },
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(horizontal = 16.dp, vertical = 8.dp),
+            singleLine = true,
+            keyboardOptions = KeyboardOptions(imeAction = ImeAction.Search),
+            keyboardActions = KeyboardActions(
+                onSearch = {
+                    onAction(CompleteSessionDetailsAction.ExecuteQuery)
+                    keyboardController?.hide()
```
aryamanshodhan: Can you explain why this is needed? Why not filter every time the user updates the query in onValueChange? I think that will help you get rid of the executeQuery state variable as well.

aryamanshodhan (changes requested):
Well done Johnny! Some changes are required because I see opportunities for optimization. Thanks!

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/components/specimens/CompleteSessionSpecimens.kt
JohnnieShen: It's just that I don't want to execute a search every time value is changed. I feel like searches should only be executed when the user presses the search button.

app/src/main/java/com/vci/vectorcamapp/complete_session/domain/model/SpecimenImageItem.kt
JohnnieShen: It's a bundle for everything we need for an image tile, it has the parent specimen, the image itself and the text that will be displayed on the badge, I think this makes it easier to pass on everything for the screen

JohnnieShen:
> Overall, it looks great! Currently, however, when I perform a search and no result appears (no specimen matches the search), and I clear the search bar, it seems like no specimen reappears. The screen continues to read "No matching specimens found." Also, maybe we can update the UI of the search bar to match the rest of the screens in a different task

You will need to also hit search after clearing the query to get the full list, and how do we want the search bar to look?

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsScreen.kt
```diff
@@ -63,7 +63,11 @@ fun CompleteSessionDetailsScreen(
 
                 CompleteSessionDetailsTab.SESSION_SPECIMENS -> CompleteSessionSpecimens(
                     session = state.session,
-                    specimensWithImagesAndInferenceResults = state.specimensWithImagesAndInferenceResults,
+                    specimenResults = state.specimensWithImagesAndInferenceResults,
```
aryamanshodhan: specimensWithImagesAndInferenceResults instead of specimenResults

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsAction.kt
```diff
@@ -5,4 +5,5 @@ import com.vci.vectorcamapp.complete_session.details.presentation.enums.Complete
 sealed interface CompleteSessionDetailsAction {
     data object ReturnToCompleteSessionListScreen : CompleteSessionDetailsAction
     data class ChangeSelectedTab(val selectedTab: CompleteSessionDetailsTab) : CompleteSessionDetailsAction
+    data class PerformSearch(val searchQuery: String) : CompleteSessionDetailsAction
```
aryamanshodhan: How are you updating the search query value without having an action dedicated to changing the value of the search field?

aryamanshodhan (changes requested):
Well done Johnny! Just a couple of minor variable renames please. Thanks!

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsAction.kt
JohnnieShen: I hold the temporary search query in the state of the screen and only pass on the query to the VM state when the user presses search. This still follows unidirectional data flow while removing the kind of duplicate searchQuery and executedSearchQuery variables in the VM state.

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/components/specimens/CompleteSessionSpecimens.kt
```diff
@@ -1,27 +1,45 @@
...
 import com.vci.vectorcamapp.ui.extensions.colors
+import com.vci.vectorcamapp.ui.extensions.dimensions
 import com.vci.vectorcamapp.ui.theme.screenWidthFraction
 
 @Composable
 fun CompleteSessionSpecimens(
     session: Session,
     specimensWithImagesAndInferenceResults: List<SpecimenWithSpecimenImagesAndInferenceResults>,
+    searchQuery: String,
+    onUpdateQuery: (String) -> Unit,
```
aryamanshodhan: onUpdateSearchQuery

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/components/specimens/CompleteSessionSpecimens.kt
```diff
@@ -31,25 +49,72 @@ fun CompleteSessionSpecimens(
             modifier = modifier.fillMaxSize()
         )
     } else {
-        LazyRow(
-            modifier = modifier.fillMaxSize(), horizontalArrangement = Arrangement.Center
+        val filteredSpecimenResults = specimensWithImagesAndInferenceResults.filter { specimenGroup ->
```
aryamanshodhan: Is there no way we can have this filtering logic be inside the view model without needing an additional list for filtered specimens?

aryamanshodhan (changes requested):
A couple of really small changes needed, I like how this is shaping up. Just a couple of questions.

aryamanshodhan (approved):

PR #96, opened 2025-08-27
SOF-250: Unit tests - Intake Screen

JohnnieShen:
This PR implements tests for the intake screen and VM. Tests include regular workflows as well as error cases. Reviewers should look through the implemented tests and if needed, propose additional tests that could be included. Reviewers should also attempt to run the tests and see if it works on their devices.


aryamanshodhan:
@JohnnieShen Can you add these tests to the CI workflows?

ThomasT0830 (changes requested):
I believe some merge conflicts need to be resolved!

PR #108, opened 2025-09-08
SOF-329: Add 1000 character text limit for session notes

aryamanshodhan:
This PR implements a very simple change involving the addition of a 1000 character limit to the session notes. No one can print the script of the bee movie into the session notes anymore (unless it is somehow compressed to < 1000 characters :) )

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
@@ -351,12 +351,14 @@ class IntakeViewModel @Inject constructor(
                 }
 
                 is IntakeAction.EnterNotes -> {
-                    _state.update {
-                        it.copy(
-                            session = it.session.copy(
-                                notes = action.text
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -351,12 +351,14 @@ class IntakeViewModel @Inject constructor(
...
-                                notes = action.text
+                    if (action.text.length <= 1000) {
+                        _state.update {
+                            it.copy(
+                                session = it.session.copy(
+                                    notes = action.text
+                                )
                             )
-                        )
+                        }
```
JohnnieShen, line 361: I think we could also use a error message for going over 1000 characters and trying to submit. Everything after 1000 is currently just sliently ignored, which might cause issues if the user assumed that it will reach the backend. Maybe an error would make it clearer. However this is a very unlikely case so take this as you will.

ThomasT0830 (approved):

PR #109, opened 2025-09-09
SOF-324: Make the arrows going back and forth clickable on Imaging Screen

JohnnieShen:
This PR makes the left and right arrow buttons clickable in imaging screeen, when clicked, it will advance the pager by 1 or go back by 1. This works because the conditions where such buttons will appear. This PR should be tested by running the app and trying if the buttons work.


ThomasT0830 (approved):
Arrows are clickable on the Motorola!

aryamanshodhan (approved):

PR #115, opened 2025-09-23
SOF-311: Sentry Logging for the IntakeScreen, RegistrationScreen, and IncompleteSessionScreen

ThomasT0830:
This pull request adds Sentry error logging to the `IntakeScreen`, `RegistrationScreen`, and `IncompleteSessionScreen` by using `Crashy.exception`. Error logging for these screens also follows the centralized Sentry error logging system set up in SOF-280. Some errors that were logged included Incomplete Session Deletion Failure (`IncompleteSessionScreen`), Session Upsert Failure (`IntakeScreen`), and Program Not Found (`RegistrationScreen`).

Testing was done on the Motorola to ensure that each error log shows up on Sentry as expected.


app/src/main/java/com/vci/vectorcamapp/registration/logging/RegistrationSentryLogger.kt
```diff
@@ -0,0 +1,49 @@
...
+            ),
+            extras = mapOf(
+                "error_context" to "Program selected by the user could not be found",
+                "recovery_action" to "User may need to re-register device",
+                "possible_causes" to "Program removed from database"
+            )
+        )
+    }
+
+    fun logUnknownError(e: Exception, failedProgramId: Int) {
```
JohnnieShen: Suggestion: Since it is currently only used in one place, I think we can rename this to something more specific.

app/src/main/java/com/vci/vectorcamapp/registration/presentation/RegistrationViewModel.kt
```diff
@@ -69,6 +71,7 @@ class RegistrationViewModel @Inject constructor(
                         _events.send(RegistrationEvent.NavigateToLandingScreen)
                     } catch (e: Exception) {
                         emitError(RegistrationError.UNKNOWN_ERROR)
+                        RegistrationSentryLogger.logUnknownError(e, selectedProgram.id)
```
JohnnieShen: Nitpick: Might want to break up this try catch block too for a little bit, since currently multiple things can go wrong and all you will get is one error for all error cases.

JohnnieShen:
Otherwise looks good! Still trying to think of high profile cases not covered already but so far nothing came to mind

aryamanshodhan (approved):
This looks good Thomas. Ready to merge. But I think we should discuss the approach to log important information that will allow us to better diagnose these issues. Additionally, the try catch blocks in our codebase can be optimized (we discussed that today).

JohnnieShen (approved):

PR #116, opened 2025-09-26
SOF-335: Add Settings Page

ThomasT0830:
This pull request adds a `SettingsScreen` that consists of an "Action" section and a "System Information" section. Each section in the `SettingsScreen` consists of a `SettingsTile` that either triggers an action or displays information. If `onClick` is passed to `SettingsTile`, the `SettingsTile` will consist of an `ActionTile` that displays a right arrow icon. If `onClick` is not passed to `SettingsTile`, the `SettingsTile` will consist of an `InfoTile` that is collapsed by default, but can be expanded to display information when clicked. A SettingsTile that displays information will have an up/down arrow instead of a right arrow icon, reflecting whether the tile is expanded or not.  Currently, `SettingsScreen` displays device information, program information, and app information, and has a `SettingsTile` that brings the user to the Intake Screen for Data Collection Mode when clicked. 

Since the "Data Collection Mode" is now only accessible through the `SettingsScreen`, the button for Data Collection Mode on the `LandingScreen` has been removed. A trailing gear icon has been added to the `ScreenHeader` of the `LandingPage` to allow users to navigate to the `SettingsScreen`.

Finally, the SplashScreen has been updated such that it no longer displays the "Version Number", as the "Version Number" can now be found on the SettingsScreen. A minor UI bug where the text on the `SplashScreen` enlarges while the `SplashScreen` is loading has been fixed.


app/src/main/java/com/vci/vectorcamapp/main/presentation/SplashScreen.kt
```diff
@@ -45,21 +49,18 @@ fun SplashScreen(
...
                 painter = painterResource(R.drawable.splash_logo),
                 contentDescription = "Splash logo",
-                modifier = Modifier.width(screenWidthFraction(0.9f))
+                modifier = Modifier
+                    .fillMaxWidth(0.8f)
+                    .wrapContentHeight()
             )
-
-            Text(
+            BasicText(
```
aryamanshodhan: What is BasicText? Why not just Text?

app/src/main/java/com/vci/vectorcamapp/main/presentation/SplashScreen.kt
```diff
@@ -45,21 +49,18 @@ fun SplashScreen(
...
-                color = MaterialTheme.colors.textPrimary,
-                modifier = Modifier.padding(vertical = MaterialTheme.dimensions.paddingSmall),
-            )
-
-            Text(
-                text = "v${BuildConfig.VERSION_NAME}",
-                style = MaterialTheme.typography.bodyLarge,
-                color = MaterialTheme.colors.textPrimary,
-                modifier = Modifier.padding(vertical = MaterialTheme.dimensions.paddingSmall),
+                style = TextStyle(
```
aryamanshodhan, line 60: We already have typography styles. Why not use one of them?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/components/SettingsActionTile.kt
```diff
@@ -0,0 +1,76 @@
...
+            if (content != null) {
+                content()
+            }
+        }
+    }
+
+    ActionTile(
+        onClick = onClick,
+        modifier = modifier,
+        content = tileContent
```
aryamanshodhan: You dont need to define a variable for tileContent. You can just do:

ActionTile(
        onClick = onClick,
        modifier = modifier
) {
     // Render whatever you want here
}

app/src/main/java/com/vci/vectorcamapp/settings/presentation/components/SettingsInfoTile.kt
```diff
@@ -0,0 +1,102 @@
...
+                    content()
+                }
+            }
+        }
+    }
+
+    ActionTile(
+        onClick = { isExpanded = !isExpanded },
+        modifier = modifier,
+        content = tileContent
```
aryamanshodhan: Same as above, you can render stuff inside the action tile component itself:

ActionTile(...) {
     // Render content here
}

aryamanshodhan (changes requested):
I am very impressed with this PR Thomas! My comments are minor nitpicks. Please look into them. Thanks, and well done!

app/src/main/java/com/vci/vectorcamapp/main/presentation/SplashScreen.kt
ThomasT0830: Unfortunately, I believe that all MaterialTheme typography styles have a default lineHeight and platformStyle settings, which (strangely) causes texts on the splash screen to expand while rendering

aryamanshodhan (approved):

PR #118, opened 2025-09-26
SOF-320: Add collection method tooltip popup window

ThomasT0830:
This pull request adds an info icon button to the `IntakeScreen` that triggers a pop-up window with explanations of the various collection methods. The info icon button is placed below the "Collection Method" dropdown, accompanied by the text "Tap to learn more about collection methods." When the info icon button or text is clicked, an `AlertDialog` titled "Collection Methods" is opened. The `AlertDialog` displays an `IntakeCollectionMethodRow` for each collection method option in the "Collection Method" dropdown that consists of an icon, a collection method title, and a collection method description. The `AlertDialog` can be dismissed by tapping on another part of the screen or by clicking on the "Done" `TextButton` in the `AlertDialog`.


app/src/main/java/com/vci/vectorcamapp/core/presentation/components/tooltip/TooltipDialog.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.font.FontWeight
+import com.vci.vectorcamapp.ui.extensions.colors
+import com.vci.vectorcamapp.ui.extensions.dimensions
+
+@Composable
+fun TooltipDialog(
+    isVisible: Boolean,
+    title: String? = null,
+    content: @Composable (() -> Unit)? = null,
```
aryamanshodhan: I always suggest having content be the last property for a composable, because I think how compose works is, the last property can be passed in using curly braces. Eg. This will enable you to do something like:

TooltipDialog(...) {
    // pass in value for last property here. If it is of type @Composable, it will allow you to render child composables inside it.
}

app/src/main/java/com/vci/vectorcamapp/core/presentation/components/tooltip/Tooltip.kt
aryamanshodhan, line 1: Is there opportunity to merge the button and Dialog into one component? I think the Dialog Row is a little overkill. The user can just pass in the Row of content directly into the content prop. If you want to abstract out the individual rows, do that at the intake screen level by creating a separate composable there. The reason I am proposing this is because it allows our tooltip to be more generic> We are not restricted to having the content be in the format of rows of icons, titles, and descriptions.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeState.kt
```diff
@@ -44,5 +44,6 @@ data class IntakeState(
         collectionDate = null,
         collectionMethod = null,
         specimenCondition = null,
-    )
+    ),
+    val showCollectionMethodTooltip: Boolean = false,
```
aryamanshodhan: isCollectionMethodTooltipVisible

aryamanshodhan (changes requested):
Good job Thomas! There are a few minor comments which I think will make this PR even more solid. Can you check out my comments please? Thanks!

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -118,6 +125,37 @@ fun IntakeScreen(
...
+                        )
+                        TooltipDialogRow(
+                            title = "Human Landing Catch",
+                            description = "A person exposes part of their body and collects mosquitoes that land on the skin.",
+                            iconPainter = painterResource(id = R.drawable.ic_human),
+                            iconDescription = "Human Landing Catch Icon"
+                        )
+                        TooltipDialogRow(
+                            title = "Pyrethrum Spray Catch",
+                            description = "A pyrethrum insecticide spray is used inside houses to knock down mosquitoes so they can be collected.",
```
JohnnieShen: @aryamanshodhan Don't forget to update this to the human written version before shipping it out

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -118,6 +125,43 @@ fun IntakeScreen(
...
+                            text = "Collection Methods",
+                            style = MaterialTheme.typography.titleLarge,
+                            color = MaterialTheme.colors.textPrimary,
+                            fontWeight = FontWeight.Bold,
+                            modifier = Modifier.padding(bottom = MaterialTheme.dimensions.paddingSmall)
+                        )
+                        CollectionMethodTooltipRow(
+                            title = "CDC Light Trap",
+                            description = "A light trap that uses a battery to attract and collect mosquitoes during the night.",
+                            iconPainter = painterResource(id = R.drawable.ic_light),
```
aryamanshodhan: I would replace this with a picture of a light trap icon. I think canva has some icons you can use. Let me know, and I can share an SVG potentially.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
aryamanshodhan: These descriptions are good. Maybe change the light trap description to: 

"A battery powered trap that uses a light to attract and collect mosquitoes at night."

aryamanshodhan (changes requested):
Looks good Thomas! One minor change needed.

aryamanshodhan (dismissed):
Well done Thomas! I think this is good to merge. Please test it once on the devices to make sure functionality is as expected. Address the merge conflict when possible please!

aryamanshodhan (dismissed):
Looks good Thomas! Thanks for addressing the conflict! Before merging this in, please make sure to thoroughly test out this PR!

aryamanshodhan (approved):

JohnnieShen (approved):
LGTM, ship it

PR #119, opened 2025-09-27
SOF-322: Make it explicit that user can go back and forth on Imaging Screen

JohnnieShen:
This PR adds a left arrow icon in imaging live preview page at the top left of the corner, it also creates a new AnimatedArrowIcon class so that it can be used in multiple places.


app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/icon/AnimatedArrowIcon.kt
aryamanshodhan, line 1: [Suggestion]: I think this component is very specific to the imaging screen, so maybe we move it to the Imaging feature. If we want to have something in core, we should consider having something a bit more generic Eg. AnimatedIcon.kt

AnimatedArrowIcon can use AnimatedIcon but still be housed inside imaging package.

aryamanshodhan (changes requested):
Well done Johnny. Just one suggestion!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/icon/AnimatedArrowIcon.kt
JohnnieShen: I'm gonna move it to imaging then, since I dont forsee us implementing more animated icons later on. Can add it back later

aryamanshodhan (dismissed):
Looks good to me! Well done Johnny!

ThomasT0830 (changes requested):
Looks great! Tested it on the Samsung A15. There is just a merge conflict that needs to be resolved!

aryamanshodhan (approved):
Well done Johnny! This is good

ThomasT0830 (approved):
This looks good to me! Tested it on the Samsung A15

PR #124, opened 2025-10-06
SOF-344: Add centroid of the mosquito marker on LiveCameraPreview

ThomasT0830:
This pull request adds a green marker to the LiveCameraPreview to mark the location of the mosquito's centroid. Some variables were also renamed in this PR for better clarity. 

The code was tested on a Samsung A15.


app/src/main/java/com/vci/vectorcamapp/ui/theme/Dimensions.kt
```diff
@@ -74,6 +74,9 @@ data class Dimensions(
 
     // Capture Animation Parameters
     val scannerLineHeight: Dp = 3.dp,
+
+    // Geometry Parameters
+    val dotRadius: Dp = 3.dp,
```
JohnnieShen: Question: What do we feel about putting this not in Dimensions but somewhere else, I don't foresee us reusing this value any time soon and I think only values that is going to be reused should be in Dimensions

JohnnieShen:
Otherwise looks good! Looks good on the Motorola

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingState.kt
```diff
@@ -29,6 +29,7 @@ data class ImagingState(
     val previewInferenceResults: List<InferenceResult> = emptyList(),
     val specimensWithImagesAndInferenceResults: List<SpecimenWithSpecimenImagesAndInferenceResults> = emptyList(),
     val focusPoint: Offset? = null,
+    val specimenCentroid: Offset? = null,
```
aryamanshodhan: Question: Is there a way to use the existing focusPoint state variable?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -175,6 +175,7 @@ class ImagingViewModel @Inject constructor(
                                 }
 
                             current.copy(
+                                specimenCentroid = liveFrameProcessingResult.specimenCentroid,
```
aryamanshodhan: Again, is this needed? Until now we achieved both autofocuses using the same focusPoint state variable.

app/src/main/java/com/vci/vectorcamapp/ui/theme/Dimensions.kt
aryamanshodhan: I agree with this!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
```diff
@@ -95,6 +98,25 @@ fun LiveCameraPreview(
                     }
                 }
         ) {
+            specimenCentroid?.let { normalized ->
```
aryamanshodhan: Why not just make it light up at the focus point? That way if there is a manual tap, we can show the marker there as well?

aryamanshodhan (changes requested):
Well done Thomas! This is clean, but there are a few questions. Can you please address the comments and resubmit a PR? Thanks!

aryamanshodhan (commented):
Question: The code is great, but I noticed something. For autofocus, everything shows up nicely - YOLO box, centroid, etc. However. as soon as the mosquito disappears from the frame, the yolo box disappears but the centroid dot remains there. Is this something we want to keep? If I do manual focus, the centroid is in the center of the circle as expected. But when I click the circle to make it disappear, the centroid point also disappears.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
```diff
@@ -100,14 +100,13 @@ fun LiveCameraPreview(
                     }
                 }
         ) {
-            focusPoint?.let { normalized ->
+            focusPoint?.takeIf { isManualFocusing || inferenceResults.isNotEmpty() }?.let { normalized ->
```
aryamanshodhan: Question: Why the takeIf? Wouldn't saying focusPoint?.let{...} be enough? Because the focus point is non-null only if we have an autofocus point or a manual focus point right?

aryamanshodhan (commented):
Hey Thomas! One Quick question. Otherwise the code looks good. I am hoping this has been thoroughly tested!

aryamanshodhan (approved):
Looks good

JohnnieShen (approved):

PR #127, opened 2025-10-10
SOF-352: Give CompleteSessionListTile a background hue based on progress

ThomasT0830:
This pull request gives the `CompleteSessionListTile` a background hue that matches the progress bar. Since the `customShadow` of the `ActionTile` previously had a fixed shadow offset, shadow color, and shadow blur, the `CompleteSessionListTile` and the `ActionTile` had to be updated to allow for such customizations.  

The code request was tested on the Samsung A15.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/complete_session/list/presentation/components/CompleteSessionListTile.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/complete_session/list/presentation/components/CompleteSessionListTile.kt
@@ -30,6 +30,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.unit.dp
 import com.vci.vectorcamapp.R
 import com.vci.vectorcamapp.core.domain.model.composites.SessionAndSite
 import com.vci.vectorcamapp.core.domain.model.helpers.SessionUploadProgress
@@ -54,6 +55,9 @@ fun CompleteSessionListTile(
     val UPLOAD_ICON_MIN_ALPHA = 0.3f
     val UPLOAD_ICON_MAX_ALPHA = 1f
 
+    val STRONG_SHADOW_ALPHA = 0.4f
+    val WEAK_SHADOW_ALPHA = 0.25f
+
     val session = sessionAndSite.session
     val site = sessionAndSite.site
     val dateFormatter = remember { SimpleDateFormat("MMM dd, yyyy", Locale.getDefault()) }
@@ -68,7 +72,11 @@ fun CompleteSessionListTile(
     session.completedAt?.let { completedAt ->
         ActionTile(
             onClick = onClick,
-            cardGlow = progressColor,
+            shadowColor = if (progressColor == MaterialTheme.colors.primary) progressColor.copy(alpha=STRONG_SHADOW_ALPHA) else progressColor.copy(alpha=WEAK_SHADOW_ALPHA),
+            shadowOffsetX = 0.dp,
+            shadowOffsetY = 0.dp,
+            shadowSpread = MaterialTheme.dimensions.shadowOffsetSmall,
+            shadowBlurRadius = MaterialTheme.dimensions.shadowBlurMedium,
             modifier = modifier,
         ) {
             Column(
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vci/vectorcamapp/complete_session/list/presentation/components/CompleteSessionListTile.kt
```diff
@@ -54,15 +55,28 @@ fun CompleteSessionListTile(
...
+    val WEAK_SHADOW_ALPHA = 0.25f
+
     val session = sessionAndSite.session
     val site = sessionAndSite.site
     val dateFormatter = remember { SimpleDateFormat("MMM dd, yyyy", Locale.getDefault()) }
     val dateTimeFormatter =
         remember { SimpleDateFormat("MMM dd, yyyy 'at' h:mm a", Locale.getDefault()) }
 
+    val progressColor =
+        if (sessionUploadProgress.totalImageCount == 0 || (sessionUploadProgress.uploadedImageCount.toFloat() / sessionUploadProgress.totalImageCount.toFloat()) == 1f) MaterialTheme.colors.primary
```
JohnnieShen: Suggestion: For this logic, I got the suggestion of using 

```
val isComplete = sessionUploadProgress.totalImageCount == 0 || sessionUploadProgress.uploadedImageCount == sessionUploadProgress.totalImageCount
```

instead on a previous PR, and I feel like it makes more sense that way, so you might consider also changing it here.

aryamanshodhan: I agree with this!

app/src/main/java/com/vci/vectorcamapp/complete_session/list/presentation/components/CompleteSessionListTile.kt
```diff
@@ -54,15 +55,28 @@ fun CompleteSessionListTile(
...
+
     session.completedAt?.let { completedAt ->
         ActionTile(
             onClick = onClick,
+            shadowColor = if (progressColor == MaterialTheme.colors.primary) progressColor.copy(alpha=STRONG_SHADOW_ALPHA) else progressColor.copy(alpha=WEAK_SHADOW_ALPHA),
+            shadowOffsetX = 0.dp,
+            shadowOffsetY = 0.dp,
+            shadowSpread = MaterialTheme.dimensions.shadowOffsetSmall,
+            shadowBlurRadius = MaterialTheme.dimensions.shadowBlurMedium,
             modifier = modifier,
```
aryamanshodhan: Have we tried out making the shadow part of the outer modifier here and removing these custom shadow paramters for the ActionTile?

app/src/main/java/com/vci/vectorcamapp/core/presentation/components/tile/ActionTile.kt
```diff
@@ -29,12 +37,12 @@ fun ActionTile(
...
-                offsetY = MaterialTheme.dimensions.shadowOffsetSmall,
-                cornerRadius = MaterialTheme.dimensions.cornerRadiusMedium,
-                spread = MaterialTheme.dimensions.shadowSpreadSmall,
-                blurRadius = MaterialTheme.dimensions.shadowBlurSmall,
+                color = shadowColor,
+                offsetX = shadowOffsetX,
+                offsetY = shadowOffsetY,
+                cornerRadius = shadowCornerRadius,
+                spread = shadowSpread,
+                blurRadius = shadowBlurRadius,
```
aryamanshodhan: Not super sure about this custom shadow parameters being passed into the action tile...Why not just have the same shadow effect for all action tiles and just pass in a "hue" attribute?

aryamanshodhan (changes requested):

aryamanshodhan (changes requested):
Question: Thoughts on having the complete session card background colors also being a very light color equal to the hue color? Eg. one tenth of the original intensity? Also, when passing in the color, dont already modify the color's intensity. Pass the original color and then modify it inside the component itself. That way you dont have to do it in multiple places.

aryamanshodhan (commented):
Thomas, thoughts on adding a hue to the entire card too? Like the background color?

aryamanshodhan (approved):

JohnnieShen (approved):

PR #132, opened 2025-10-21
SOF-367: Replace all instances of loading animation with Splash Screen

JohnnieShen:
This PR replaced all instances of the legacy loading animation with the new splash screen. The old loading animation implementation is still in the codebase but is no longer used


aryamanshodhan (changes requested):
Hey Johnny! The changes are good. However, LoadingAnimation is no longer used anywhere so you should delete the LoadingAnimation.ts file. There are also some dead imports which need to be removed. Make sure there are no instances of LoadingAnimation anywhere in the code.

JohnnieShen:
I was gonna leave it there in case we want to revive it later but sure

aryamanshodhan (approved):
Looks good! This is ready to merge.

ThomasT0830 (approved):
Looks great on the Samsung A15!

PR #133, opened 2025-10-22
SOF-351: Placeholder number in intake form doesn't feel intuitive

JohnnieShen:
This PR implements a fix for an issue where some users reported that the numbers in the intake form doesn't feel intuitive as they are displayed as placeholders but are treated as real values. This PR makes it so that the values will be filled as real value from the beginning and no placeholder will be involved.


app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -498,8 +519,20 @@ fun IntakeScreen(
 
                     TextEntryField(
                         label = "Number of LLINs Available",
-                        value = if (surveillanceForm.numLlinsAvailable == 0) "" else surveillanceForm.numLlinsAvailable.toString(),
-                        onValueChange = { onAction(IntakeAction.EnterNumLlinsAvailable(it.filter { character -> character.isDigit() })) },
+                        value = (surveillanceForm.numLlinsAvailable ?: 0).toString(),
```
aryamanshodhan: Warning: Elvis operator (?:) always returns the left operand of non-nullable type Int

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -498,8 +519,20 @@ fun IntakeScreen(
...
+                        onValueChange = { newValue ->
+                            val oldValue = (surveillanceForm.numLlinsAvailable ?: 0).toString()
```
aryamanshodhan: Warning: Elvis operator (?:) always returns the left operand of non-nullable type Int

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -396,8 +396,20 @@ fun IntakeScreen(
...
+                        value = surveillanceForm.numPeopleSleptInHouse.toString(),
+                        onValueChange = { newValue ->
+                            val oldValue = surveillanceForm.numPeopleSleptInHouse.toString()
+                            val filteredNewValue = newValue.filter { it.isDigit() }
+
+                            val finalValueString = if (oldValue == "0" && filteredNewValue.length > 1) {
+                                filteredNewValue.filter { it != '0' }
+                            } else {
+                                filteredNewValue
+                            }
```
aryamanshodhan: Since this logic became a lot more complicated, it has to move into the view model. This has to be done for all other similar code.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -396,8 +396,20 @@ fun IntakeScreen(
...
+
+                            val normalizedFinalValue = finalValueString.toIntOrNull()?.toString() ?: "0"
```
aryamanshodhan: Question: Should we keep the value zero by default or the empty string? I am in favor of the latter option because otherwise the user might mistakenly submit zero for the numbers.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -264,23 +261,19 @@ class IntakeViewModel @Inject constructor(
...
+                        )
                     }
                 }
 
                 is IntakeAction.EnterNumLlinsAvailable -> {
-                    val numLlinsAvailable =
-                        if (action.count.isBlank()) 0 else action.count.toIntOrNull()
-                    numLlinsAvailable?.let { count ->
+                    val numLlinsAvailable = action.count.toIntOrNull() ?: 0
+                    numLlinsAvailable.let { count ->
```
aryamanshodhan: The let is not needed here because numLlinsAvailable will never be null

aryamanshodhan (changes requested):
Nice one Johnny! Needs a few edits but great job!

app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateMonthsSinceIrsUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.vci.vectorcamapp.intake.domain.use_cases
+
+import com.vci.vectorcamapp.core.domain.util.Result
+import com.vci.vectorcamapp.intake.domain.util.FormValidationError
+import javax.inject.Inject
+
+class ValidateMonthsSinceIrsUseCase @Inject constructor() {
+    operator fun invoke(count: Int): Result<Unit, FormValidationError> {
+        return if (count <= -1) {
```
aryamanshodhan, line 9: Make sure this is consistent across all use cases. In some it says <= -1. In others it says == -1.

app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateNumLlinsAvailableUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.vci.vectorcamapp.intake.domain.use_cases
+
+import com.vci.vectorcamapp.core.domain.util.Result
+import com.vci.vectorcamapp.intake.domain.util.FormValidationError
+import javax.inject.Inject
+
+class ValidateNumLlinsAvailableUseCase @Inject constructor() {
+    operator fun invoke(count: Int): Result<Unit, FormValidationError> {
+        return if (count <= -1) {
```
aryamanshodhan, line 9: Make sure this is consistent across all use cases. In some it says <= -1. In others it says == -1.

app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateNumPeopleSleptInHouseUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.vci.vectorcamapp.intake.domain.use_cases
+
+import com.vci.vectorcamapp.core.domain.util.Result
+import com.vci.vectorcamapp.intake.domain.util.FormValidationError
+import javax.inject.Inject
+
+class ValidateNumPeopleSleptInHouseUseCase @Inject constructor() {
+    operator fun invoke(count: Int): Result<Unit, FormValidationError> {
+        return if (count == -1) {
```
aryamanshodhan: Make sure this is consistent across all use cases. In some it says <= -1. In others it says == -1.

app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateNumPeopleSleptUnderLlinUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.vci.vectorcamapp.intake.domain.use_cases
+
+import com.vci.vectorcamapp.core.domain.util.Result
+import com.vci.vectorcamapp.intake.domain.util.FormValidationError
+import javax.inject.Inject
+
+class ValidateNumPeopleSleptUnderLlinUseCase @Inject constructor() {
+    operator fun invoke(count: Int): Result<Unit, FormValidationError> {
+        return if (count == -1) {
```
aryamanshodhan: Make sure this is consistent across all use cases. In some it says <= -1. In others it says == -1.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -496,45 +487,29 @@ fun IntakeScreen(
                         })
 
                     if (surveillanceForm.wasIrsConducted) {
```
aryamanshodhan: I think this if condition is not required anymore if you are using 

surveillanceForm.monthsSinceIrs?.let

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -496,45 +487,29 @@ fun IntakeScreen(
...
-                                val normalizedFinalValue = finalValueString.toIntOrNull()?.toString() ?: "0"
-                                onAction(IntakeAction.EnterMonthsSinceIrs(normalizedFinalValue))
-                            },
-                            placeholder = "0",
-                            singleLine = true,
-                        )
+                        surveillanceForm.monthsSinceIrs?.let {
+                            TextEntryField(
+                                label = "Months Since IRS",
+                                value = if (it < 0)
```
aryamanshodhan: Instead of "it", can we use monthsSinceIrs?

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeScreen.kt
```diff
@@ -576,22 +551,13 @@ fun IntakeScreen(
                     surveillanceForm.numPeopleSleptUnderLlin?.let { current ->
```
aryamanshodhan: current is unused. Can we replace it with numPeopleSleptUnderLlin, and then replace accordingly in line 554?

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -116,6 +116,30 @@ class IntakeViewModel @Inject constructor(
...
+                            intakeValidationUseCases.validateNumLlinsAvailable(it.numLlinsAvailable)
+                        }
+
+                    val numPeopleSleptUnderLlinResult =
+                        if (surveillanceForm?.numPeopleSleptUnderLlin != null) {
+                            intakeValidationUseCases.validateNumPeopleSleptUnderLlin(surveillanceForm.numPeopleSleptUnderLlin)
+                        } else {
+                            null
+                        }
+
```
aryamanshodhan, line 138: Thoughts on simplifying to:

val numPeopleSleptInHouseResult =
    surveillanceForm?.let {
        intakeValidationUseCases.validateNumPeopleSleptInHouse(it.numPeopleSleptInHouse)
    }

val monthsSinceIrsResult =
    surveillanceForm?.takeIf { it.wasIrsConducted && it.monthsSinceIrs != null }?.let {
        intakeValidationUseCases.validateMonthsSinceIrs(it.monthsSinceIrs!!)
    }

val numLlinsAvailableResult =
    surveillanceForm?.let {
        intakeValidationUseCases.validateNumLlinsAvailable(it.numLlinsAvailable)
    }

val numPeopleSleptUnderLlinResult =
    surveillanceForm?.numPeopleSleptUnderLlin?.let {
        intakeValidationUseCases.validateNumPeopleSleptUnderLlin(it)
    }

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
aryamanshodhan, line 99: If we are defining

val session = _state.value.session
val surveillanceForm = _state.value.surveillanceForm

Lets maybe use it throughout this function instead of writing _state.value.session and _state.value.surveillanceForm everywhere

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -271,34 +307,35 @@ class IntakeViewModel @Inject constructor(
...
+                    val normalized = normalizeNumericInput(oldValue, action.count)
+                    val numLlinsAvailable = normalized.toIntOrNull() ?: 0
+                    _state.update {
+                        it.copy(
+                            surveillanceForm = it.surveillanceForm?.copy(
+                                numLlinsAvailable = numLlinsAvailable
+                            )
+                        )
+                    }
+                    if (numLlinsAvailable == 0) {
```
aryamanshodhan: Shouldn't this be <= 0? Otherwise it will show the values even if this field is -1 (empty string)? Is this condition handled somewhere else which I am missing?

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -563,4 +603,17 @@ class IntakeViewModel @Inject constructor(
...
+        val oldValueString = (oldValue ?: 0).toString()
+        val filteredNewValue = newValue.filter { it.isDigit() }
+
+        val finalValueString = if (oldValueString == "0" && filteredNewValue.length > 1) {
+                filteredNewValue.filter { it != '0' }
+            } else {
+                filteredNewValue
+            }
+
+        return finalValueString.toIntOrNull().toString()
```
aryamanshodhan, line 623: What if this returns the string value "null"?

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -563,4 +603,17 @@ class IntakeViewModel @Inject constructor(
             }
         }
     }
+
+    private fun normalizeNumericInput(oldValue: Int?, newValue: String): String {
```
aryamanshodhan, line 613: Can you explain this function please?

aryamanshodhan (changes requested):
This is very close to done Johnny. Just a few final changes and questions remaining. Also I noticed, when I press backspace, it resets the value to 0 instead of the empty string (-1). Is this something we want?

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
JohnnieShen: It is exactly the logic I had elsewhere, just transplanted here. If the original value is 0 and we type in an extra number before or after it then it removes the zero. that's it

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
JohnnieShen: This will not happen, old value will start at null but new value is not nullable. Since we derive finalValueString from newValue, the null case will not happen

aryamanshodhan: Also quick question Johnny. 

finalValueString.toIntOrNull().toString()

Why not just return finalValueString?

Why are we converting the string to int and then back to string, especially since we are already filtering out the non-digit characters?

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
```diff
@@ -122,10 +122,8 @@ class IntakeViewModel @Inject constructor(
                         }
 
                     val monthsSinceIrsResult =
-                        if (surveillanceForm?.wasIrsConducted == true && surveillanceForm.monthsSinceIrs != null) {
-                            intakeValidationUseCases.validateMonthsSinceIrs(surveillanceForm.monthsSinceIrs)
-                        } else {
-                            null
+                        surveillanceForm?.takeIf { it.wasIrsConducted && it.monthsSinceIrs != null }?.let {
+                            intakeValidationUseCases.validateMonthsSinceIrs(it.monthsSinceIrs!!)
```
aryamanshodhan: [nitpick]: Using "!!" always scares me, even thought the takeIf guards against this. I would recommend changing this to:

val monthsSinceIrsResult =
                        surveillanceForm?.takeIf { it.wasIrsConducted && it.monthsSinceIrs != null }?.let {
                            it.monthsSinceIrs?.let { months -> intakeValidationUseCases.validateMonthsSinceIrs(months) }
                        }

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
aryamanshodhan: Question here. If normalized is guaranteed to be a string of digits (because we filter out non-digits in the function), so we need toIntOrNull? I appreciate and prefer the defensive programming, but let me know what you think. Also, when deleting a digit, what is your thought on resetting to -1 (blank) instead of 0?

aryamanshodhan (changes requested):
This is basically ready Johnny. I have a couple of questions.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
JohnnieShen: I left it in there for formatting reasons, in case finalValueString doesn't look like a integer, personally I think it doesn't hurt to leave it in there just to ensure that the format is good

app/src/main/java/com/vci/vectorcamapp/intake/presentation/IntakeViewModel.kt
JohnnieShen: I still feel like leaving out a bit of extra formatting guards is a good idea, let me experiemnt with resetting to -1 and i will get back to you

aryamanshodhan (approved):
Looks good Johnny!

ThomasT0830 (approved):
Looks good to me!

PR #134, opened 2025-10-22
SOF-163: Compress locally stored images after successful upload 

JohnnieShen:
This PR implements functionality for compressing the locally stored copy of specimen images after successful upload. The local images will be compressed with 40 quality, for images originally around ~3MB, the size will be ~300MB post compression. The value can be tuned if user require higher definition or even smaller size.


app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
```diff
@@ -327,6 +331,19 @@ class ImageUploadWorker @AssistedInject constructor(
...
+                Log.d(
+                    "ImageUploadWorker",
+                    "Upload successful. Compressing original file: ${task.image.imageUri}"
+                )
+                compressImageFile(task.image.imageUri, COMPRESSION_QUALITY)
+            } catch (e: Exception) {
+                Log.e(
+                    "ImageUploadWorker",
+                    "Failed to compress original file (${task.image.imageUri}) after upload.",
+                    e
```
aryamanshodhan, line 351: Question: Should we update the notification? Is this considered failure?

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
```diff
@@ -558,4 +575,26 @@ class ImageUploadWorker @AssistedInject constructor(
...
+            resolver.openOutputStream(uri, "wt")?.use { out ->
+                if (!bitmap.compress(Bitmap.CompressFormat.JPEG, quality, out)) {
+                    throw IOException("Failed to compress bitmap to $uri")
+                }
+                out.flush()
+            } ?: throw IOException("Failed to open output stream for $uri")
+
+            Log.d("ImageUploadWorker", "Successfully compressed $uri to $quality% quality.")
+        } finally {
+            bitmap?.recycle()
```
aryamanshodhan, line 632: Question: Is recycling the bitmap necessary? Isn't android's garbage collection handling this by itself?

aryamanshodhan (commented):
Well done Johnny! This is good, there are a couple of questions. I am also gonna verify the compression percentage with Sunny before I approve this PR.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
JohnnieShen: I don't think it should be considered a failure, since the file made it to the backend so the upload task itself succeeded (and this logic would only run after that is successful) we can notify the user somehow (or not) but the task itself should be considered successful

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
JohnnieShen: Good question, I can look into it but doesn't hurt to manually handle garbage collection when we can

aryamanshodhan: Do we have an update on this?

JohnnieShen: https://stackoverflow.com/questions/18452192/is-it-needed-to-call-bitmap-recycle-after-used-in-android

it's not necesssary in newer versions of Android as GC handles it, but is often recommended as a best practice to free up memory when the bitmap is no longer needed.

Also next time please also ping me in Slack if i don't respond! My github notifications have been broken for a long time now so I don't get updates from comment chains.

aryamanshodhan (commented):
This looks good to me, but I was the last pusher. Someone else should test again and approve this PR.

aryamanshodhan (changes requested):
This is good Johnny, but there is a merge conflict. Keep the null value as default instead of 0L

aryamanshodhan (commented):
Looks good Johnny! 

Can we do another round of testing for this PR before we merge it in? I am worried about data loss when compressing the images. Can you provide a quick couple of sentences about the tests you have already conducted, and the outcomes?

JohnnieShen:
Tested following scenarios:

Image uploads fail before it starts -> Image not compressed
Image upload fails during upload -> Image not compressed
Image upload concludes but finish fails -> image not compressed
Image upload concludes and finishes and uploader gets the return result -> Image compressed

The following scenario wasn't originally considered but is now patched:
Image upload concludes, image compression fail with some exception
Original implementation: Since we use "wt", file will be truncated and we will be left with a corrupted image
New implementation: We save a temp copy of the original image, in the case of exception, we restore the original image, afterwards we delete the temp file once we are sure compression is successful

ThomasT0830 (dismissed):
Looks good on the A32!

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
```diff
@@ -558,4 +580,52 @@ class ImageUploadWorker @AssistedInject constructor(
...
         notificationManager.notify(notificationId, notification)
     }
+
+    private suspend fun compressImageFile(uri: Uri, quality: Int) = withContext(Dispatchers.IO) {
+        val resolver = context.contentResolver
+        var bitmap: Bitmap? = null
+        var backupFile: File? = null
+
+        try {
+            backupFile = File.createTempFile("img_backup_", ".tmp", context.cacheDir)
```
aryamanshodhan: Question: Will this file naming cause the same issue that we observed with the data upload? Eg. some images not getting compressed

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
```diff
@@ -558,4 +580,52 @@ class ImageUploadWorker @AssistedInject constructor(
...
+            Log.d("ImageUploadWorker", "Successfully compressed $uri to $quality% quality.")
+        } catch (e: Exception) {
+            Log.e("ImageUploadWorker", "Error compressing $uri, attempting to restore original", e)
+            if (backupFile != null && backupFile.exists()) {
+                try {
+                    resolver.openOutputStream(uri, "wt")?.use { out ->
+                        backupFile.inputStream().use { backupIn ->
+                            backupIn.copyTo(out)
+                        }
+                        out.flush()
```
aryamanshodhan, line 623: Can you explain what this is doing during the exception branch?

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
```diff
@@ -558,4 +580,52 @@ class ImageUploadWorker @AssistedInject constructor(
...
+                    }
+                    Log.d("ImageUploadWorker", "Successfully restored original image from backup for $uri")
+                } catch (restore: Exception) {
```
aryamanshodhan, line 626: What is restore? In this case should we be deleting the backup file? Or will it do it in the other finally block?

aryamanshodhan (commented):
Well done Johnny! I have some questions about the backupFile stuff.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
JohnnieShen: Yeah that is a probability, I can fix it.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
JohnnieShen: If there is an exception during compression, i.e. IOException from bitmap.compress for whatever reason, we stop the compression workflow, and if the temporary backup file exists we use "wt" to truncate the original file (which at this moment have been partially written into and is essentially corrupted so we can disregard its content), and write the backup file's content (uncompressed original image) back into the original file.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
JohnnieShen: It will be done in the finally block in backupfile.delete.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
aryamanshodhan: If the backup file is corrupted, why should we write its contents to the original uncompressed file? I am not sure I understand. Can you also explain again the purpose of the backup file and what edge case you are trying to address with it?

JohnnieShen: 1. If the backup file is corrupted, why should we write its contents to the original uncompressed file? 

In the exception case it is the original file (which is the file that is compressed) that gets compressed. Backup shouldn't corrupt in that case and in most cases, well if backup case is corrupted then tough luck.

2. Can you also explain again the purpose of the backup file

The backup file's content is the data of the original uncompressed image

3. what edge case you are trying to address with it? 

e.g. bitmap.compress throws IOException. To simulate:

resolver.openOutputStream(uri, "wt")?.use { out ->
                throw IOException("TEST EXCEPTION") // ADD THIS LINE HERE
                if (!bitmap.compress(Bitmap.CompressFormat.JPEG, quality, out)) {
                    throw IOException("Failed to compress bitmap to $uri")
                }
                out.flush()
            } ?: throw IOException("Failed to open output stream for $uri")

aryamanshodhan: Got it. Valid points. Last question. What happens if there is an exception during the compression? It will just move onto the next image and not compress it right? If that is the case, is that acceptable?

JohnnieShen: in the original implementation: it will termiate the compression process. No other consequences, but since the stream is terminated the rest of the content of the compressed file will not be written into the file, the compressed image will be incomplete and at many times corrupted. Since we use "wt" it truncates the original file before writing in so it will destroy the original content

JohnnieShen: In the new implementation, yeah, that

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
```diff
@@ -558,4 +580,57 @@ class ImageUploadWorker @AssistedInject constructor(
...
+            backupFile = File.createTempFile("img_backup_${specimenId}_${imageId}", ".tmp", context.cacheDir)
+
+            resolver.openInputStream(uri)?.use { input ->
+                FileOutputStream(backupFile).use { out ->
+                    input.copyTo(out)
+                }
+            } ?: throw IOException("Failed to open input stream for $uri to create backup")
+
+            bitmap = backupFile.inputStream().use { backupIn ->
+                BitmapFactory.decodeStream(backupIn)
```
aryamanshodhan, line 604: Just make sure that running bitmap operations does not lead to changing MD5 numbers of the images. For example, if you have a bitmap which you compress to 100 quality JPEG, which in turn you compress to 100 quality JPEG, even though the qualities are 100, they will lead to different MD5 checksums for both images. So just make sure that the backup file is identical to the original file, both in case of successful and failed compression.

aryamanshodhan (approved):
I have approved this PR Johnny. Tested it on the A15, and all looks good. Can you please test this once again on the motorola before merging it in? Also, I left a note for you. Please read it. Thanks!

app/src/main/java/com/vci/vectorcamapp/core/data/upload/image/ImageUploadWorker.kt
JohnnieShen: Sure, will do in a bit

ThomasT0830 (approved):
Looks good to me!

PR #136, opened 2025-10-22
SOF-370: Registration screen background image is out of view

aryamanshodhan:
This PR involves a change to the registration screen background image. With the addition of collector name and title on the screen, the mosquito in the original image gets hidden. This PR replaces that image so that it looks nicer.


JohnnieShen (approved):
LGTM as long as we are fine with AI art

ThomasT0830 (approved):

PR #137, opened 2025-10-25
SOF-384: Add new column to specimen table - wasSelectedForFurtherProcessing

JohnnieShen:
This PR adds an additional field to the specimen class and all corresponding classes called wasSelectedForFurtherProcessing, this field will be used to track whether a specimen is selected for further processing and will be used in UI to indicate that to the user.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
@@ -6,5 +6,6 @@ import kotlinx.serialization.Serializable
 data class SpecimenDto(
     val id: Int? = null,
     val specimenId: String = "",
-    val sessionId: Int = -1
+    val sessionId: Int = -1,
+    val wasSelectedForFurtherProcessing: Boolean? = null
 )
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
@@ -7,5 +7,5 @@ data class SpecimenDto(
     val id: Int? = null,
     val specimenId: String = "",
     val sessionId: Int = -1,
-    val wasSelectedForFurtherProcessing: Boolean? = null
+    val wasSelectedForFurtherProcessing: Boolean = false
 )
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
@@ -7,5 +7,5 @@ data class SpecimenDto(
     val id: Int? = null,
     val specimenId: String = "",
     val sessionId: Int = -1,
-    val wasSelectedForFurtherProcessing: Boolean = false
+    val shouldProcessFurther: Boolean = false
 )
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_15_16_ADD_SPECIMEN_SELECTED_FOR_FURTHER_PROCESSING_COLUMN.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_15_16_ADD_SPECIMEN_SELECTED_FOR_FURTHER_PROCESSING_COLUMN.kt
@@ -1,0 +1,12 @@
+package com.vci.vectorcamapp.core.data.room.migrations.versions
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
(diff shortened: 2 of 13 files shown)


app/src/main/java/com/vci/vectorcamapp/core/data/dto/specimen/SpecimenDto.kt
```diff
@@ -6,5 +6,6 @@ import kotlinx.serialization.Serializable
 data class SpecimenDto(
     val id: Int? = null,
     val specimenId: String = "",
-    val sessionId: Int = -1
+    val sessionId: Int = -1,
+    val wasSelectedForFurtherProcessing: Boolean? = null
```
aryamanshodhan: Make this a required boolean. Either true or false. It will help avoid the need for handling null values. Default can be false.

app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_15_16_ADD_SPECIMEN_SELECTED_FOR_FURTHER_PROCESSING_COLUMN.kt
```diff
@@ -0,0 +1,12 @@
+package com.vci.vectorcamapp.core.data.room.migrations.versions
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+val Migration_15_16_ADD_SPECIMEN_SELECTED_FOR_FURTHER_PROCESSING_COLUMN = object : Migration(15, 16){
```
aryamanshodhan: Suggestion: Rename so that MIGRATION is also all caps.

app/src/main/java/com/vci/vectorcamapp/core/domain/model/Specimen.kt
```diff
@@ -2,5 +2,6 @@ package com.vci.vectorcamapp.core.domain.model
 
 data class Specimen(
     val id: String,
-    val remoteId: Int?
+    val remoteId: Int?,
+    val wasSelectedForFurtherProcessing: Boolean = false
```
aryamanshodhan: For domain models, we should not use default values. This will help avoid the issue where we accidentally forget to define a property. In this case android studio will highlight the issue with an error. You should remove the default value and update the instances of Specimen throughout the codebase to help account for the creation of new specimens. For now, you can set all instances to false. When we implement the randomization algorithm, we can change to true based on the probability.

aryamanshodhan (changes requested):
Great job Johnny! Just some minor changes needed. Thanks for the quick turnaround!

app/src/main/java/com/vci/vectorcamapp/core/data/upload/metadata/MetadataUploadWorker.kt
```diff
@@ -436,7 +436,7 @@ class MetadataUploadWorker @AssistedInject constructor(
                 }
 
             val remoteSpecimen = Specimen(
-                id = remoteSpecimenDto.specimenId, remoteId = remoteSpecimenDto.id
+                id = remoteSpecimenDto.specimenId, remoteId = remoteSpecimenDto.id, wasSelectedForFurtherProcessing = false
```
aryamanshodhan: wasSelectedForFurtherProcessing = remoteSpecimenDto.wasSelectedForFurtherProcessing

aryamanshodhan (changes requested):
Last change needed. Well done Johnny!

aryamanshodhan (changes requested):
The backend called the variable shouldProcessFurther. Can we rename instances of wasSelectedForFurtherProcessing to shouldProcessFurther?

app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_15_16_ADD_SPECIMEN_SELECTED_FOR_FURTHER_PROCESSING_COLUMN.kt
```diff
@@ -0,0 +1,12 @@
+package com.vci.vectorcamapp.core.data.room.migrations.versions
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+val MIGRATION_15_16_ADD_SPECIMEN_SELECTED_FOR_FURTHER_PROCESSING_COLUMN = object : Migration(15, 16){
```
aryamanshodhan: nitpick: I think the migration variable name should also be updated. I also noticed that the file name needs to be changed to Camel case, not all caps. Check out how other migration files are named.

aryamanshodhan (dismissed):
Optional Comment left.

ThomasT0830 (dismissed):
Looks good to me!

aryamanshodhan (changes requested):
There is a bug - the backend right now never receives the shouldProcessFurther field from the app.

ThomasT0830 (dismissed):
Looks good to me! Tested it on the Samsung A32

aryamanshodhan (approved):

PR #139, opened 2025-10-29
SOF-350: Make Specimen ID text field on Imaging Screen red when error related to specimen ID is emitted

JohnnieShen:
This PR implements a functionality where the text entry field in imaging screen will turn the text red upon emitting an error related to specimen ID validation.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/core/presentation/components/form/TextEntryField.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/presentation/components/form/TextEntryField.kt
@@ -34,6 +34,8 @@ fun TextEntryField(
     keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
     keyboardActions: KeyboardActions = KeyboardActions.Default,
     maxCharacters: Int = 200,
+    showErrorMessage: Boolean = true,
+    changeTextColorOnError: Boolean = false,
 ) {
     val context = LocalContext.current
 
@@ -73,6 +75,12 @@ fun TextEntryField(
                 }
             },
             maxLines = 8,
+            textStyle = MaterialTheme.typography.bodyMedium.copy(
+                color = if (error != null && changeTextColorOnError)
+                    MaterialTheme.colors.error
+                else
+                    MaterialTheme.colors.textPrimary
+            ),
             colors = OutlinedTextFieldDefaults.colors(
                 focusedBorderColor = MaterialTheme.colors.transparent,
                 unfocusedBorderColor = MaterialTheme.colors.transparent,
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vci/vectorcamapp/core/presentation/components/form/TextEntryField.kt
```diff
@@ -34,6 +34,8 @@ fun TextEntryField(
     keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
     keyboardActions: KeyboardActions = KeyboardActions.Default,
     maxCharacters: Int = 200,
+    showErrorMessage: Boolean = true,
+    changeTextColorOnError: Boolean = false,
```
aryamanshodhan: Suggestion: Lets change the text color all the time. No need to pass it in as a parameter. I think error != null is enough to determine whether we need to color the text red or not.

app/src/main/java/com/vci/vectorcamapp/core/presentation/components/form/TextEntryField.kt
```diff
@@ -97,7 +105,7 @@ fun TextEntryField(
                 .heightIn(min = MaterialTheme.dimensions.componentHeightMedium)
         )
 
-        if (error != null) {
+        if (error != null && showErrorMessage) {
```
aryamanshodhan, line 107: Question: This is good, but is there a way we can refactor the code so that we don't need the showErrorMessage flag?

aryamanshodhan (changes requested):
Looks good Johnny, there are a couple of suggestions.

app/src/main/java/com/vci/vectorcamapp/core/presentation/components/form/TextEntryField.kt
JohnnieShen: Sure, I wasn't sure whether we want to since it would change how the intake screen look as well but we can do that

app/src/main/java/com/vci/vectorcamapp/core/presentation/components/form/TextEntryField.kt
JohnnieShen: I am not sure if I follow, since we need to control whether we need to display the error text in different occasions don't we need this flag?

aryamanshodhan: I guess my question was, is the error parameter enough for us to determine whether to show the message or not? Or do we definitely need a separate flag to keep track of that?

JohnnieShen: I added it in there because there is the case here where there is an error but you don't want to display the error text, only make the text red. So we will need a seperate flag even though we have an error parameter already

aryamanshodhan (approved):
Looks good to me

ThomasT0830 (approved):

PR #144, opened 2025-11-04
SOF-382: Add alert dialog for selection mosquitoes at random for further processing

ThomasT0830:
This pull request implements a mandatory packaging confirmation dialog that appears when specimens are randomly selected for additional laboratory processing. When a user is saving a specimen image, the app checks if further processing is required (`shouldProcessFurther`) and if packaging has not been completed yet (`hasConfirmedPackaging`). If both conditions are true, a non-dismissible alert dialog appears requiring the user to acknowledge that they have to package the specimen before proceeding. The dialog is non-dismissible, as it prevents dismissal through outside clicks or back button presses. The dialog also includes a checkbox that must be checked and a continue button that remains disabled until confirmation is provided. When the user dismisses the dialog after confirming packaging, the original save action is automatically retried, allowing the normal save flow to proceed. This approach prevents the user from accidentally skipping the packaging step for specimens that require further laboratory analysis. 

The pull request has been tested on the Motorola.


app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -288,6 +288,77 @@ fun ImagingScreen(
                     )
                 }
 
+                if (state.showProcessFurtherDialog) {
+                    AlertDialog(
+                        onDismissRequest = { },
```
aryamanshodhan, line 294: Question: Should we dismiss the dialog here? If this allows dismissal without acknowledgement, then we should not.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -296,10 +305,28 @@ class ImagingViewModel @Inject constructor(
...
+                    _state.update {
+                        it.copy(currentSpecimen = it.currentSpecimen.copy(shouldProcessFurther = true))
+                    }
+
+                    val shouldProcessFurther = _state.value.currentSpecimen.shouldProcessFurther
+
+                    if (shouldProcessFurther && !_state.value.hasConfirmedPackaging) {
+                        _state.update { it.copy(showProcessFurtherDialog = true) }
+                        return@launch
+                    }
```
aryamanshodhan, line 324: You can just leave everything false for now. I will go ahead and make the changes when I implement the rest of the task. Write the code assuming that the randomization will work. I looked at the UI and it looks great!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingViewModel.kt
```diff
@@ -296,10 +305,28 @@ class ImagingViewModel @Inject constructor(
...
                     }
 
+                    // TODO: Delete after business logic is implemented
+                    _state.update {
+                        it.copy(
+                            currentSpecimen = it.currentSpecimen.copy(shouldProcessFurther = true)
+                        )
+                    }
+
+                    _state.update {
```
aryamanshodhan: Duplicate code

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -288,6 +288,77 @@ fun ImagingScreen(
...
+                                Text(
+                                    text = "Specimen Selected for Further Processing",
+                                    color = MaterialTheme.colors.icon
+                                )
+                            }
+                        },
+                        text = {
+                            Column {
+                                Text(
+                                    text = "This specimen has been randomly selected and must be packed separately for additional laboratory processing. Please package this specimen for further processing now before continuing.",
```
aryamanshodhan: This is a lot of text which is saying the same thing multiple times. Maybe compress it a bit, make it more succinct.

aryamanshodhan (dismissed):
Great job Thomas! This needs a couple of small changes, but looks good otherwise!

JohnnieShen (approved):

ThomasT0830 (commented):
Everything looks good to me!

PR #165, opened 2025-12-16
SOF-399: Fix model discrepancy between A32 and Motorola

ThomasT0830:
This pull request attempts to resolve any model discrepancy between the A32 and the Motorola. Previously, there was an issue in which the model logits differed between the Samsung A32 and the Motorola, and model performance dropped significantly on the Samsung A32. This was because Samsung devices and Android devices use different image decoders. Therefore, the pull request addresses the issue by ensuring that all devices use OpenCV's image decoder, thereby standardizing the image decoder across devices. 

While the fix has been implemented and tested on the A32, additional testing is needed to confirm that the model discrepancy among the Samsung A32, the Motorola, and the PC is resolved.


aryamanshodhan (approved):
Approving conditional on the fact that PC outputs match phone outputs

JohnnieShen (approved):
^^^ I agree

PR #167, opened 2025-12-21
SOF-406: Native camera resolution during image capture

JohnnieShen:
This pull request refactors the camera integration in the imaging module to use lower-level CameraX APIs and Compose-based preview, replacing the previous use of `LifecycleCameraController` and `PreviewView`. The changes improve flexibility, enable more direct control over camera use cases, and modernize the UI layer by leveraging `CameraXViewfinder` for live preview. The refactor also updates how image capture and focus are handled throughout the codebase.

**Camera integration and UI modernization:**
- Replaced `LifecycleCameraController` and `PreviewView` with direct use of CameraX use cases (`Preview`, `ImageCapture`, `ImageAnalysis`) and Compose's `CameraXViewfinder` for rendering the camera preview. This change allows more granular control over camera behavior and better integration with Jetpack Compose. [[1]](diffhunk://#diff-11da20e3054bcf862de03eb30f8770056c03a74d9224972d82ec8ad0a3e9aff2R90-R163) [[2]](diffhunk://#diff-11da20e3054bcf862de03eb30f8770056c03a74d9224972d82ec8ad0a3e9aff2L551-R622) [[3]](diffhunk://#diff-c42c1e41059e3e561092771d258695fa91ec39dff24f0f8802d5fd785fc545a4L3-R6) [[4]](diffhunk://#diff-c42c1e41059e3e561092771d258695fa91ec39dff24f0f8802d5fd785fc545a4L33-R34) [[5]](diffhunk://#diff-c42c1e41059e3e561092771d258695fa91ec39dff24f0f8802d5fd785fc545a4L42-R63) [[6]](diffhunk://#diff-c42c1e41059e3e561092771d258695fa91ec39dff24f0f8802d5fd785fc545a4L70-R81)

**Image capture flow updates:**
- Updated all usages of image capture to accept an `ImageCapture` instance instead of a `LifecycleCameraController`. This includes changes to the repository interface and implementation, view model, and UI actions, ensuring consistency and compatibility with the new camera setup. [[1]](diffhunk://#diff-166a14192013b0aa838dd70dfb043a13b2f23c9b52404821184875ba2fddef34R4-R12) [[2]](diffhunk://#diff-dbae17508cfea5303bd485830eb7d24e6294a6ffea635558c2a8fd4d5a0e3cc3L32-R36) [[3]](diffhunk://#diff-c1bc24975ae1eb79f74f0da93553e774b5d2a513daf664ccfa5027e1bdca8cd4L271-R271) [[4]](diffhunk://#diff-1a6ca5ac1bceced4bda273800ddefe921b2c065e92db837e16f582ac1795a790L13-R13) [[5]](diffhunk://#diff-11da20e3054bcf862de03eb30f8770056c03a74d9224972d82ec8ad0a3e9aff2L586-R661)

**Focus and metering improvements:**
- Refactored the focus controller to use `CameraControl` and `CameraInfo` directly, with a `DisplayOrientedMeteringPointFactory` for accurate focus point calculation. This enables more precise and reliable focus behavior, aligned with the new camera setup.

**Dependency updates:**
- Added the `androidx.camera.compose` dependency to support the new Compose-based camera preview.

**Minor code and import cleanups:**
- Removed unused imports and legacy code related to the old camera controller and preview setup, streamlining the codebase for maintainability. [[1]](diffhunk://#diff-dbae17508cfea5303bd485830eb7d24e6294a6ffea635558c2a8fd4d5a0e3cc3R8) [[2]](diffhunk://#diff-11da20e3054bcf862de03eb30f8770056c03a74d9224972d82ec8ad0a3e9aff2R49-R56) [[3]](diffhunk://#diff-11da20e3054bcf862de03eb30f8770056c03a74d9224972d82ec8ad0a3e9aff2L62-R79) [[4]](diffhunk://#diff-c42c1e41059e3e561092771d258695fa91ec39dff24f0f8802d5fd785fc545a4L19-L24)

These changes collectively modernize the camera integration, improve code clarity, and lay the groundwork for future enhancements to the imaging workflow.


app/build.gradle.kts
```diff
@@ -147,6 +147,7 @@ dependencies {
     implementation(libs.androidx.camera.video)
     implementation(libs.androidx.camera.mlkit.vision) // CameraX ML Kit Vision Integration
     implementation(libs.androidx.camera.extensions) // CameraX Extensions library
+    implementation(libs.androidx.camera.compose) // CameraX Compose integration
```
aryamanshodhan, line 149: Question: Now that we are using this camera compose library, are there any redundant dependencies we can get rid of?

app/src/main/java/com/vci/vectorcamapp/imaging/data/camera/CameraFocusControllerImplementation.kt
```diff
@@ -1,59 +1,47 @@
...
 
-class CameraFocusControllerImplementation (
-    private val previewView: PreviewView,
-    private val controller: LifecycleCameraController,
+class CameraFocusControllerImplementation(
+    private val cameraControl: CameraControl?,
+    private val cameraInfo: CameraInfo?,
+    private val display: Display,
+    private val width: Float,
+    private val height: Float
```
aryamanshodhan, line 19: Can you explain each of these parameters? Why did we remove the PreviewView?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
```diff
@@ -30,7 +30,8 @@ import com.vci.vectorcamapp.ui.extensions.dimensions
 
 @Composable
 fun LiveCameraPreview(
-    controller: LifecycleCameraController,
+    surfaceRequest: SurfaceRequest?,
+    camera: Camera?,
```
aryamanshodhan, line 34: Question: Why are each of these nullable?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
```diff
@@ -39,44 +40,45 @@ fun LiveCameraPreview(
...
+        val containerWidth = with(density) { maxWidth.toPx() }
+        val containerHeight = with(density) { maxHeight.toPx() }
+        val containerSize = IntSize(containerWidth.toInt(), containerHeight.toInt())
 
-        AndroidView(factory = { previewView }, modifier = Modifier.fillMaxSize())
+        val cameraFocusController = remember(camera, containerWidth, containerHeight) {
+            CameraFocusControllerImplementation(
+                cameraControl = camera?.cameraControl,
+                cameraInfo = camera?.cameraInfo,
+                display = view.display,
```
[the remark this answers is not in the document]
aryamanshodhan: This sounds problematic. Please look into this comment and address the issue. Maybe there is a video posted by Philip Lackner as to how to use this new approach.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
```diff
@@ -39,44 +40,45 @@ fun LiveCameraPreview(
...
         modifier = modifier
             .fillMaxWidth()
             .aspectRatio(1f / MaterialTheme.dimensions.aspectRatio)
     ) {
-        val containerSize = IntSize(
-            width = with(density) { maxWidth.roundToPx() },
-            height = with(density) { maxHeight.roundToPx() })
+        val containerWidth = with(density) { maxWidth.toPx() }
+        val containerHeight = with(density) { maxHeight.toPx() }
+        val containerSize = IntSize(containerWidth.toInt(), containerHeight.toInt())
```
aryamanshodhan, line 55: Question: How is this different from what we had before?

val containerSize = IntSize(
            width = with(density) { maxWidth.roundToPx() },
            height = with(density) { maxHeight.roundToPx() }
)

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
```diff
@@ -39,44 +40,45 @@ fun LiveCameraPreview(
...
+                cameraFocusController.cancelFocus()
+            } else {
+                cameraFocusController.focusAt(focusPoint)
+            }
+        }
+
+        if (surfaceRequest != null) {
+            CameraXViewfinder(
+                surfaceRequest = surfaceRequest,
+                implementationMode = ImplementationMode.EXTERNAL,
```
aryamanshodhan, line 78: Question: Are we sure ImplementationMode.EXTERNAL is the one we want? Are there other implementation modes?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -78,19 +87,80 @@ fun ImagingScreen(
...
-        LifecycleCameraController(context).apply {
-            setEnabledUseCases(CameraController.IMAGE_CAPTURE or CameraController.IMAGE_ANALYSIS)
-            setImageAnalysisAnalyzer(ContextCompat.getMainExecutor(context), analyzer)
-            imageCaptureFlashMode = ImageCapture.FLASH_MODE_OFF
-            imageCaptureResolutionSelector = ResolutionSelector.Builder()
-                .setResolutionStrategy(ResolutionStrategy.HIGHEST_AVAILABLE_STRATEGY).build()
+    val view = LocalView.current
+
+    LaunchedEffect(Unit) {
+        val rotation = view.display.rotation
```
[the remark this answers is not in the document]
aryamanshodhan: Sounds problematic. Please look into this, and clean up the code.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -78,19 +87,80 @@ fun ImagingScreen(
...
+
+    val isReviewing by rememberUpdatedState(newValue = state.currentImageBytes != null)
     val analyzer = remember {
         SpecimenImageAnalyzer { frame ->
-            onAction(ImagingAction.ProcessFrame(frame))
+            if (!isReviewing) {
+                onAction(ImagingAction.ProcessFrame(frame))
+            } else {
+                frame.close()
+            }
```
aryamanshodhan, line 102: Question: Can you explain the purpose of isReviewing? I thought the analyzer and view model are already accounting to see whether processing is happening, and if it is, then the frame is directly closed. Not sure if we need both. Also, was wondering why not just use 

if (state.currentImageBytes == null) {
                onAction(ImagingAction.ProcessFrame(frame))
            } else {
                frame.close()
            }

Why do we need rememberUpdatedState?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -78,19 +87,80 @@ fun ImagingScreen(
 
     val scope = rememberCoroutineScope()
 
+    var surfaceRequest by remember { mutableStateOf<SurfaceRequest?>(null) }
+    var imageCaptureUseCase by remember { mutableStateOf<ImageCapture?>(null) }
+    var camera by remember { mutableStateOf<Camera?>(null) }
```
aryamanshodhan, line 93: Is there a way we can directly initialize these 3 values like we had done for the camera controller initially? That way we don't need to handle null cases and worry about null pointer exceptions.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -78,19 +87,80 @@ fun ImagingScreen(
...
+
+        val imageCapture = ImageCapture.Builder()
+            .setCaptureMode(ImageCapture.CAPTURE_MODE_MAXIMIZE_QUALITY)
+            .setTargetRotation(rotation)
+            .setResolutionSelector(
+                ResolutionSelector.Builder()
+                    .setAspectRatioStrategy(AspectRatioStrategy.RATIO_4_3_FALLBACK_AUTO_STRATEGY)
+                    .setAllowedResolutionMode(ResolutionSelector.PREFER_HIGHER_RESOLUTION_OVER_CAPTURE_RATE)
+                    .build()
+            )
```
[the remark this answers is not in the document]
aryamanshodhan: Follow up to this:
1. To make sure that we use minimal inference time, make sure that the preview has lower resolution (I think the current one we use is 640 by 480). 
2. To make sure we get highest accuracy, make sure the capture of the image uses the highest quality.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -78,19 +87,80 @@ fun ImagingScreen(
...
+            .build()
+
+        val imageAnalysis = ImageAnalysis.Builder()
+            .setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
+            .setTargetRotation(rotation)
+            .setResolutionSelector(
+                ResolutionSelector.Builder()
+                    .setResolutionStrategy(ResolutionStrategy.HIGHEST_AVAILABLE_STRATEGY)
```
aryamanshodhan: Make sure analysis is not using too large of a resolution, otherwise the YOLO box will take time to update.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -78,19 +87,80 @@ fun ImagingScreen(
...
+                    .build()
+            )
+            .build()
+
+        imageAnalysis.setAnalyzer(ContextCompat.getMainExecutor(context), analyzer)
+
+        try {
```
aryamanshodhan, line 151: Should we be binding inside the LiveCameraPreview? Maybe we can pass in the use cases? Lets stick to the exact architecture we have currently to avoid unexpected lifecycle related issues. Eg. if the controller was passed into the live camera preview page and bound inside that component, lets move the use cases and binding there because we replaced the LifecycleCameraController with the individual use cases. Should this binding process live inside a LaunchedEffect? How did we bind the controller before?

aryamanshodhan: Treat (ImageCapture + ImageAnalysis) as the new controller, and wherever we passed in controller, we now need to pass in the 2 use cases.

Is this correct? You have done more reading into the documentation about this. Is this how it works?

aryamanshodhan (changes requested):
Hey Johnny! This is great! There are a few questions I have about this implementation, but overall seems quite clean. Please address the comments ASAP, and do more testing on this.

app/src/main/java/com/vci/vectorcamapp/imaging/data/camera/CameraFocusControllerImplementation.kt
JohnnieShen: Well because we are using the new camera X view instead of the more abstracted PreviewView. cameraInfo gives us the sensor's orientation relative to the device, display gives the device's current orientation and width and height are used to normalize tap coordinates

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
JohnnieShen: This is because we also need to use containerWidth / height later so we also put them into their own variables

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
JohnnieShen: Because I think initialization of CameraX is async, so there will be a period of time which the camera is being started and during then Camera and SurfaceRequest is not available yet

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/LiveCameraPreview.kt
JohnnieShen: Implementation modes (performance vs. composition)
CameraXViewfinder can render the preview in two ways:

EXTERNAL (SurfaceView-backed)
Camera frames render on their own Surface composed by the system outside your Compose draw pass. Think “a live video layer behind your UI.”. Often enables hardware overlays → best performance/latency. Great for full-screen rectangular previews behind standard UI chrome. Per-pixel effects on the camera pixels (complex clipping/blur) won’t apply because it’s a separate layer.

Pros: lower latency, less GPU work, great for full-screen preview/recording.
Cons: not affected by per-pixel UI effects (rounded masks/blur), won’t show up in Compose screenshots.

EMBEDDED (TextureView-backed)
Camera frames as a GPU texture drawn inside your Compose render pass — like a redrawable panel, behaving like any other composable. You get deep clipping/masking/animations/blur/z-ordering, at the cost of more GPU work and slightly higher latency.

Pros: behaves like normal UI; clips, alpha, blur, fancy shapes, complex z-order all work.
Cons: more GPU work → slightly higher latency/jank risk under heavy UI or on mid-range devices.

https://www.droidcon.com/2025/11/05/compose-native-camerax-is-now-stable-end-to-end-guide-for-jetpack-compose/

I believe EXTERNAL is what we would want here

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
JohnnieShen: Because in val analyzer = remember { ... it only runs in the initial composition when currentImageBytes are null, so if user takes image this block is not re-run and the old analyzer is returned and the original state is used. By using a rememberUpdatedState we avoid this issue and would evaluate the new state every recomposition

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
JohnnieShen: I think the current way of doing this is not a bad idea because conceptually camera and surfacerequest won't be initialized at the beginning and need to wait to be initialized, so initial null is a good representation of this.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
JohnnieShen: Yes, ImageCapture and Analysis and Camera should be treated as controller. I am thinking if we should move the binding but I think we should leave it as is. Moving the binding to liveCameraPreview would bind camera lifecycle to the UI component which might be not ideal. Since if the Ui component is no longer in the composition the camera might unbind and be rebinded later which is expensive. In the old code the binding logic with LifecycleCameraController is also in ImagingScreen too so we are maintaining the same pattern

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
aryamanshodhan: In that case, do we need to use the isProcessing state variable? Can we stick to just using rememberUpdatedState? Maybe we remove isProcessing from the state and rename isReviewing to isProcessing?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -139,7 +140,6 @@ fun ImagingScreen(
             .setTargetRotation(rotation)
             .setResolutionSelector(
                 ResolutionSelector.Builder()
-                    .setResolutionStrategy(ResolutionStrategy.HIGHEST_AVAILABLE_STRATEGY)
                     .build()
```
aryamanshodhan, line 145: In order to remain consistent, lets add the aspect ratio strategy here as well.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
aryamanshodhan, line 115: Lets also set the target rotation here to be consistent

aryamanshodhan, line 164: Can we add a new task to the backlog to handle this error gracefully? I think the least we should do is reset the camera, image capture use case, and surface request to null.

aryamanshodhan (changes requested):
I think this is good Johnny. Last couple of things need tweaks then I think this can be merged.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
JohnnieShen: I could swear I thought I did it ...

aryamanshodhan (approved):
Looks good Johnny! Well done

ThomasT0830 (approved):
Looks good to me! Tested the code on the A32!

PR #170, opened 2026-01-30
SOF-413: Create new entities and models for location hierarchies

ThomasT0830:
This pull request introduces support for hierarchical location types in the app's data model and database, allowing for more flexible and normalized representation of site locations. It adds a new `LocationType` entity, updates the `Site` model and related DTOs/entities to include hierarchy fields, and ensures that UI components handle nullable location data gracefully. The database schema and migrations are updated to support these new fields and tables.

**Database schema and migration updates:**
- Added a new `LocationTypeEntity` and corresponding `location_type` table, with foreign key to `ProgramEntity`, and registered new migrations for schema changes. 
- Updated `SiteEntity` to include new fields: `locationTypeId`, `parentId`, `name`, and `locationHierarchy`, made location fields nullable, and added corresponding indices. Registered new migration to add these columns and indices.

**Repository and DAO changes:**
- Added `LocationTypeDao` and `LocationTypeRepositoryImplementation` for CRUD operations on location types. 
- Registered `LocationTypeDao` in `VectorCamDatabase`. 

**Data model, DTO, and mapping updates:**
- Updated `SiteDto`, `SiteEntity`, and domain models to support new hierarchy fields and nullable location fields. Updated all related mappers for these changes.
- Added GSON library dependency for JSON serialization/deserialization.

**UI improvements:**
- Updated UI components to safely display site location details only if they are present (null-safe rendering), reflecting the new nullable data model. 

**Other:**
- Minor cleanup in `SessionRepositoryImplementation` by removing unnecessary exception printing.


app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_22_23MakeSiteLocationColumnsNullable.kt
aryamanshodhan, line 1: This file needs to be renamed. There is no _ between 23 and MakeSiteLocationColumnsNullable

app/src/main/java/com/vci/vectorcamapp/core/data/util/HierarchySerializer.kt
aryamanshodhan, line 1: It seems like this file is not used anywhere.

Additionally, I dont think we need this file. If necessary, we should instead write a type converter to convert the entity to and from JSON. Like how we did it for FloatListConverter, SessiontypeConverter, etc. And then add that converter inside VectorCamDatabase.kt

app/src/main/java/com/vci/vectorcamapp/core/data/util/HierarchySerializer.kt
```diff
@@ -0,0 +1,23 @@
+package com.vci.vectorcamapp.core.data.util
+
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+
+object HierarchySerializer {
```
aryamanshodhan: LocationHierarchySerializer

app/build.gradle.kts
```diff
@@ -223,6 +223,9 @@ dependencies {
     androidTestImplementation(libs.androidx.ui.test.junit4) // Compose JUnit testing
     debugImplementation(libs.androidx.ui.tooling) // Debugging tools for Compose
     debugImplementation(libs.androidx.ui.test.manifest) // Debugging Compose manifest tests
+
+    // GSON Library
+    implementation(libs.gson) // For serialization and deserialization of JSON files
```
aryamanshodhan: Question: Why do we need the GSON library? Does KotlinX serialization not work for this?

app/src/main/java/com/vci/vectorcamapp/core/domain/model/Site.kt
```diff
@@ -2,11 +2,15 @@ package com.vci.vectorcamapp.core.domain.model
...
+    val subCounty: String? = null,
+    val parish: String? = null,
+    val villageName: String? = null,
+    val houseNumber: String? = null,
+    val healthCenter: String? = null,
+    val isActive: Boolean,
+    val locationTypeId: Int? = null,
+    val parentId: Int? = null,
+    val name: String? = null,
+    val locationHierarchy: String? = null
```
aryamanshodhan: Shouldn't this be a JSON object?

app/src/main/java/com/vci/vectorcamapp/core/data/dto/site/SiteDto.kt
aryamanshodhan, line 1: SiteDto was updated but the RoomDatabaseModule was not updated to seed the database correctly. Maybe that can be part of the next task where we dynamically seed based on program selection, but just flagging this here.

app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_22_23MakeSiteLocationColumnsNullable.kt
aryamanshodhan, line 1: Explain this migration file and the SQL please. Also format it according to how the other migration files are formatted with the back ticks around column names and primary key/foreign key definitions at the end of the table.

app/src/main/java/com/vci/vectorcamapp/core/data/mappers/SiteMapper.kt
aryamanshodhan, line 1: I think in order to remain consistent with other mappers, we should remove the foreign key attributes from the domain (NOT THE ENTITY). This includes locationTypeId and parentId.

app/src/main/java/com/vci/vectorcamapp/core/domain/model/Site.kt
```diff
@@ -2,11 +2,15 @@ package com.vci.vectorcamapp.core.domain.model
...
-    val isActive: Boolean
+    val district: String? = null,
+    val subCounty: String? = null,
+    val parish: String? = null,
+    val villageName: String? = null,
+    val houseNumber: String? = null,
+    val healthCenter: String? = null,
+    val isActive: Boolean,
+    val locationTypeId: Int? = null,
+    val parentId: Int? = null,
```
aryamanshodhan: I think locationTypeId and parentId should not be part of the Site domain model to remain consistent with other domain models, since they are foreign keys. In the toEntity function, we should take in a locationTypeId and parentId as function arguments (like how we have programId as one of the arguments into that function).

app/src/main/java/com/vci/vectorcamapp/core/data/room/entities/SiteEntity.kt
```diff
@@ -12,16 +12,25 @@ import androidx.room.PrimaryKey
         childColumns = ["programId"],
         onDelete = ForeignKey.CASCADE,
         onUpdate = ForeignKey.CASCADE
-    )], indices = [Index("programId")]
+    )],
```
aryamanshodhan: We should also add foreign keys here:
1. Foreign key locationTypeId should reference LocationType "id" field.
2. Self-referencing foreign key where parentId references Site "id" field.

aryamanshodhan (changes requested):
Well done Johnny and Thomas. I think we need to add some modifications to this implementation before it can be merged in. Also, I am not clear in this implementation where the JSON to and from string conversion happens before data is saved to the RoomDb, or retrieved from the roomDB. Someone please clarify this. Thanks!

app/build.gradle.kts
aryamanshodhan, line 1: Since the file using this new library is not used anywhere, maybe we can get rid of this dependency altogether? Please remove it from this gradle file, as well as libs.versions.toml

aryamanshodhan (changes requested):

app/src/main/java/com/vci/vectorcamapp/core/data/room/migrations/versions/Migration_22_23MakeSiteLocationColumnsNullable.kt
ThomasT0830: The purpose of this file is to make some columns for Site nullable. However, we cannot change the nullability directly in SQL, so we basically have to copy the entire table to a new table with the columns now nullable, drop the old table, and rename the new table to match the old table.

aryamanshodhan (approved):

JohnnieShen (approved):

PR #188, opened 2026-03-07
feat(SOF-423): include location hierarchy to be searchable

stephen1706:
unnamed (1)

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
@@ -64,6 +64,11 @@ class CompleteSessionDetailsViewModel @Inject constructor(
                             add(imageAndInferenceResult.specimenImage.abdomenStatus)
                             add(imageAndInferenceResult.specimenImage.metadataUploadStatus.displayText(context))
                             add(imageAndInferenceResult.specimenImage.imageUploadStatus.displayText(context))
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
```diff
@@ -64,6 +64,11 @@ class CompleteSessionDetailsViewModel @Inject constructor(
                             add(imageAndInferenceResult.specimenImage.abdomenStatus)
                             add(imageAndInferenceResult.specimenImage.metadataUploadStatus.displayText(context))
                             add(imageAndInferenceResult.specimenImage.imageUploadStatus.displayText(context))
+                            _state.value.site.locationHierarchy?.values?.forEach {
```
aryamanshodhan: This screen just displays a list of specimens, and they will all be taken in the same location because it was from the same session. Therefore, I don't think we need the search filter in this file.

aryamanshodhan (commented):
Thank you so much Stephen! This looks great! I think there is one place where the location filtering is not needed, because we display a list of specimens and if they are part of the same session, they are all taken in the same location.

app/src/main/java/com/vci/vectorcamapp/complete_session/details/presentation/CompleteSessionDetailsViewModel.kt
stephen1706: okay i have reverted this change

aryamanshodhan (approved):

PR #190, opened 2026-03-13
fix: use openCV rotation

stephen1706:
Rotate image using open CV method

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/extensions/ImageProxyExtensions.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/extensions/ImageProxyExtensions.kt
@@ -2,6 +2,7 @@ package com.vci.vectorcamapp.imaging.presentation.extensions
 
 import android.graphics.Bitmap
 import androidx.camera.core.ImageProxy
+import androidx.core.graphics.createBitmap
 import org.opencv.android.Utils.bitmapToMat
 import org.opencv.android.Utils.matToBitmap
 import org.opencv.core.Core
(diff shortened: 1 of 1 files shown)


ThomasT0830 (changes requested):
Hi Stephen! The PR looks great! I tested the code on the Motorola and found that it significantly reduced the runtime of `toUprightBitmap()`. I also verified that using OpenCV to perform the rotation did not cause any discrepancy in model output between the Motorola and the PC. 

This is something super minor, but Android Studio is showing a warning that says, "Use the KTX function createBitmap instead" for the line `val result = Bitmap.createBitmap(dst.cols(), dst.rows(), Bitmap.Config.ARGB_8888)`. Besides that, the code looks good to me. Thank you!

stephen1706:
> Hi Stephen! The PR looks great! I tested the code on the Motorola and found that it significantly reduced the runtime of `toUprightBitmap()`. I also verified that using OpenCV to perform the rotation did not cause any discrepancy in model output between the Motorola and the PC.
> 
> This is something super minor, but Android Studio is showing a warning that says, "Use the KTX function createBitmap instead" for the line `val result = Bitmap.createBitmap(dst.cols(), dst.rows(), Bitmap.Config.ARGB_8888)`. Besides that, the code looks good to me. Thank you!

@ThomasT0830 i pushed an update to clear the warning, can you please try again on motorola as i don't have the test device. thanks!

ThomasT0830 (approved):
Looks good to me! Tested the code on the Motorola!

aryamanshodhan (approved):

PR #191, opened 2026-03-16
feat(SOF-449):  add camera metadata

stephen1706:
add camera metadata when capturing image

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/network/api/RemoteSpecimenImageDataSource.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/network/api/RemoteSpecimenImageDataSource.kt
@@ -2,6 +2,7 @@ package com.vci.vectorcamapp.core.data.network.api
 
 import com.vci.vectorcamapp.BuildConfig
 import com.vci.vectorcamapp.core.data.dto.inference_result.InferenceResultDto
+import com.vci.vectorcamapp.core.data.dto.specimen_image.ImageMetadataDto
 import com.vci.vectorcamapp.core.data.dto.specimen_image.PostSpecimenImageRequestDto
 import com.vci.vectorcamapp.core.data.dto.specimen_image.PostSpecimenImageResponseDto
 import com.vci.vectorcamapp.core.data.dto.specimen_image.SpecimenImageDto
@@ -19,6 +20,7 @@ import io.ktor.client.request.post
 import io.ktor.client.request.setBody
 import io.ktor.http.ContentType
 import io.ktor.http.contentType
+import kotlinx.serialization.json.Json
 import javax.inject.Inject
 
 class RemoteSpecimenImageDataSource @Inject constructor(
@@ -54,6 +56,12 @@ class RemoteSpecimenImageDataSource @Inject constructor(
                                 sexInferenceDuration = it.sexInferenceDuration,
                                 abdomenStatusInferenceDuration = it.abdomenStatusInferenceDuration
                             )
+                        },
+                        imageMetadata = specimenImage.imageMetadata?.let {
+                            @Suppress("SwallowedException")
+                            try { 
+                                Json.decodeFromString<ImageMetadataDto>(it) 
+                            } catch (_: Exception) { null }
                         }
                     )
                 )
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/network/api/RemoteSpecimenImageDataSource.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/network/api/RemoteSpecimenImageDataSource.kt
@@ -56,7 +56,7 @@ class RemoteSpecimenImageDataSource @Inject constructor(
                                 abdomenStatusInferenceDuration = it.abdomenStatusInferenceDuration
                             )
                         },
-                        imageMetadata = specimenImage.imageMetadata?.toImageMetadataDto()
+                        metadata = specimenImage.imageMetadata?.toImageMetadataDto()
                     )
                 )
             }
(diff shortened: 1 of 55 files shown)


app/src/main/java/com/vci/vectorcamapp/core/domain/model/SpecimenImage.kt
```diff
@@ -13,5 +13,6 @@ data class SpecimenImage(
     val metadataUploadStatus: UploadStatus,
     val imageUploadStatus: UploadStatus,
     val capturedAt: Long,
-    val submittedAt: Long?
+    val submittedAt: Long?,
+    val imageMetadata: String? = null
```
aryamanshodhan: Maybe we remove default null value from the metadata to ensure that we provide a value every time the specimen image is instantiated.

JohnnieShen (dismissed):
lGTM

ThomasT0830 (changes requested):
Hi! I just have a few questions about the PR. I'll leave some comments here. Thanks!

app/src/main/java/com/vci/vectorcamapp/imaging/domain/model/CameraMetadata.kt
ThomasT0830, line 1: I have a quick question! Would it make sense for CameraMetadata to be in something like imaging/domain/model/ instead of presentation/? Since it is capturing raw Camera2 sensor data, it feels more like a domain model than a UI state. Thank you!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
```diff
@@ -138,7 +145,45 @@ fun ImagingScreen(
                     .setAllowedResolutionMode(ResolutionSelector.PREFER_HIGHER_RESOLUTION_OVER_CAPTURE_RATE)
                     .build()
             )
-            .build()
+
+        Camera2Interop.Extender(imageCaptureBuilder)
```
ThomasT0830, line 145: I was also wondering whether it would make sense to extract this Camera2 capture callback part into its own class, similar to how we have CameraFocusController for focus logic? I am not really sure if this capture result parsing fits in the Composable. Thank you!

ThomasT0830, line 145: I am not really familiar with this, but I was just wondering if there a way to get the capture metadata without going through the experimental Camera2 Interop? Thanks!

app/src/main/java/com/vci/vectorcamapp/core/data/network/api/RemoteSpecimenImageDataSource.kt
```diff
@@ -54,6 +56,12 @@ class RemoteSpecimenImageDataSource @Inject constructor(
                                 sexInferenceDuration = it.sexInferenceDuration,
                                 abdomenStatusInferenceDuration = it.abdomenStatusInferenceDuration
                             )
+                        },
```
ThomasT0830, line 58: Should we handle deserialization of imageMetadata the same way we do for other JSON fields from the backend (e.g., location hierarchy)? If I understand correctly, since ImageMetadataDto is annotated with @Serializable, Ktor should handle deserialization automatically without needing the Json.decodeFromString call? Thank you!

ThomasT0830 (dismissed):
Hi Stephen! I tested the PR on the Motorola and it works great! I just had a few questions that I was not sure about. I would really appreciate it if you could take a look when you get a chance! Thank you so much!

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/ImagingScreen.kt
stephen1706: @ExperimentalCamera2Interop has been in CameraX since 1.0 alpha (2020) and hasn't had a breaking API change. The @Experimental marker just means Google reserves the right to change it, it doesn't mean it's unstable or risky in practice. The only real alternative to get all seven fields would be to bypass CameraX entirely and use the raw Camera2 API, which will broke other existing features

stephen1706: extracted the logic into outside LaunchedEffect

app/src/main/java/com/vci/vectorcamapp/imaging/domain/model/CameraMetadata.kt
stephen1706: moved this file into imaging/domain/model/ instead

app/src/main/java/com/vci/vectorcamapp/core/data/network/api/RemoteSpecimenImageDataSource.kt
stephen1706: updated to use imageMetadata directly instead of json string

app/src/main/java/com/vci/vectorcamapp/core/domain/model/SpecimenImage.kt
```diff
@@ -13,5 +14,6 @@ data class SpecimenImage(
     val metadataUploadStatus: UploadStatus,
     val imageUploadStatus: UploadStatus,
     val capturedAt: Long,
-    val submittedAt: Long?
+    val submittedAt: Long?,
+    val imageMetadata: CameraMetadata? = null
```
aryamanshodhan: Question: Should we remove the null value? When a domain model is instantiated, it would be nice to explicitly set the values of the fields at that time.

aryamanshodhan (commented):
Looks good Stephen! Just one comment.

app/src/main/java/com/vci/vectorcamapp/core/domain/model/SpecimenImage.kt
stephen1706: code updated to avoid default value

app/src/main/java/com/vci/vectorcamapp/core/domain/model/SpecimenImage.kt
stephen1706: code updated to avoid default value

aryamanshodhan (approved):

PR #193, opened 2026-03-28
SOF-447: Settings screen should include resync button to be able to re-pull updated list of the location types, sites

JohnnieShen:
This pull request adds a new "Resync Locations" feature to the Settings screen, allowing users to manually trigger a sync of location data. The implementation includes UI changes, state management, error handling, and the backend logic for fetching and updating location types and sites. Additionally, a new error type and user-facing error message are introduced for sync failures.

**New Feature: Location Resync**

- Added a "Resync Locations" action to the Settings screen, including UI feedback (progress indicator) while syncing is in progress. [[1]](diffhunk://#diff-57c2a7ecb134b8873f8eb1363cb7e618c664bbc59958626e211957ca25172b9eR71-R94) [[2]](diffhunk://#diff-09f98b8032f29bc834f4ad97d209a84f4008324cbd235df8d996d0b2b418e48dL29-R30) [[3]](diffhunk://#diff-cc51df79e91ce9715e4f4c01fb8af8dd58783a793328b3237184460e9b7c3b91R18)
- Implemented `resyncLocations()` in `SettingsViewModel`, which fetches location types and sites from remote sources, updates the local database, and handles related session/cache updates. [[1]](diffhunk://#diff-11ac9a9c800bbe6febaab82b6e1368878a1b90bb656ccf711e0d5afe7b17e1c2R202-R294) [[2]](diffhunk://#diff-11ac9a9c800bbe6febaab82b6e1368878a1b90bb656ccf711e0d5afe7b17e1c2R4-R19) [[3]](diffhunk://#diff-11ac9a9c800bbe6febaab82b6e1368878a1b90bb656ccf711e0d5afe7b17e1c2R31) [[4]](diffhunk://#diff-11ac9a9c800bbe6febaab82b6e1368878a1b90bb656ccf711e0d5afe7b17e1c2R46-R58)

**Error Handling and User Feedback**

- Introduced a new `SettingsError.LOCATION_SYNC_FAILED` error type and mapped it to a new string resource for user-friendly error messages. [[1]](diffhunk://#diff-479241b67519e1c21d30fa14d26e6c42c3d4556201989847177b101fb2a51eb0L7-R8) [[2]](diffhunk://#diff-29f77924e9dcd053b2a34a5cc827fdc65443da0918704030b5207d27f74f2999R126) [[3]](diffhunk://#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103R220-R222)

**UI and State Management**

- Updated `SettingsState` to include an `isSyncingLocations` flag for UI state tracking during sync operations.
- Added a progress indicator to the UI to provide feedback during the sync process. [[1]](diffhunk://#diff-57c2a7ecb134b8873f8eb1363cb7e618c664bbc59958626e211957ca25172b9eR14) [[2]](diffhunk://#diff-57c2a7ecb134b8873f8eb1363cb7e618c664bbc59958626e211957ca25172b9eR71-R94)

These changes collectively improve the reliability and user experience of managing location data within the app.


app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsAction.kt
```diff
@@ -16,6 +16,7 @@ interface SettingsAction {
     data object ShowDeleteCollectorDialog : SettingsAction
     data object DismissDeleteCollectorDialog : SettingsAction
     data object ConfirmDeleteCollector : SettingsAction
+    data object ResyncData : SettingsAction
```
aryamanshodhan: Suggestion: Maybe rename this to ResyncProgramData?

app/src/main/res/values/strings.xml
```diff
@@ -220,6 +220,9 @@
     <!-- Settings error: Collector deletion failed -->
     <string name="settings_error_collector_deletion_failed">Could not delete collector. Please try again.</string>
 
+    <!-- Settings error: Data sync failed -->
+    <string name="settings_error_data_sync_failed">Could not sync data. Please try again.</string>
```
aryamanshodhan, line 224: Question: Is this something that gets displayed if I try to sync without internet connection?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,7 +260,129 @@ class SettingsViewModel @Inject constructor(
...
                     }
                 }
+                SettingsAction.ResyncData -> {
+                    resyncData()
+                }
+            }
+        }
+    }
+
+    private suspend fun resyncData() {
```
aryamanshodhan, line 286: I would suggest handling this function like how we have it in RegistrationViewModel. Sync location types, then sites, then form, then questions. We already have mappers for all the data types. You should use those. Check out how this is done in that view model as well. Obviously there are a couple of additional considerations when resyncing from the settings screen vs. fetching for the first time in the registration screen, which you might need to handle...but I suggest using the registration vie model functions as a starting point for that. Additionally, using the code:
(... as Result.success).data is not good practice. Handle it as separate success and error branches inside a "when" statement.
Also, why are we setting the default intake fields cache for the resyncing?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -32,22 +50,43 @@ class SettingsViewModel @Inject constructor(
...
+    private val formRepository: FormRepository,
+    private val formQuestionRepository: FormQuestionRepository,
+    private val defaultIntakeFieldsCache: DefaultIntakeFieldsCache,
+    private val currentSessionCache: CurrentSessionCache,
+    connectivityObserver: ConnectivityObserver,
     errorMessageEmitter: ErrorMessageEmitter,
 ) : CoreViewModel(errorMessageEmitter) {
 
+    @Inject
+    lateinit var transactionHelper: TransactionHelper
```
aryamanshodhan: Question: Shouldn't this be injected into the view model directly instead of lateinit? How is this done elsewhere?

app/src/main/java/com/vci/vectorcamapp/core/data/mappers/FormQuestionMapper.kt
```diff
@@ -50,7 +50,7 @@ fun FormQuestionDto.toDomain(): FormQuestion {
     )
 }
 
-private fun FormQuestionPrerequisiteExpressionDto.toDomain(): FormQuestionPrerequisiteExpression {
+fun FormQuestionPrerequisiteExpressionDto.toDomain(): FormQuestionPrerequisiteExpression {
```
aryamanshodhan: Question: Why do we need to make this a public function? We didn't need to do that for the registration view model?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsScreen.kt
```diff
@@ -68,11 +69,41 @@ fun SettingsScreen(
...
                     onClick = { onAction(SettingsAction.StartNewPracticeSession) },
                     modifier = modifier
                 )
+
+                Row(
+                    modifier = Modifier.fillMaxWidth(),
+                    horizontalArrangement = Arrangement.SpaceBetween,
+                    verticalAlignment = Alignment.CenterVertically
+                ) {
+                    SettingsActionTile(
```
aryamanshodhan: Question: Instead of making this an action tile, can we make it a separate section within the screen? Maybe with a little section that becomes green when we have internet, and red text when there is no internet?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/components/SettingsActionTile.kt
```diff
@@ -26,10 +27,18 @@ import com.vci.vectorcamapp.ui.extensions.dimensions
 fun SettingsActionTile(
     title: String,
     modifier: Modifier = Modifier,
+    enabled: Boolean = true,
     onClick: (() -> Unit),
     content: (@Composable ColumnScope.() -> Unit)? = null,
 ) {
-    ActionTile(onClick = onClick, modifier = modifier) {
+    ActionTile(
```
aryamanshodhan, line 32: Is we make the resync button a separate section of the settings screen, we might not need to modify this file.

aryamanshodhan (changes requested):
Hey Johnny! This is a good start. There are a few things I would suggest changing, especially the way the resync function is set up right now.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: Are we not supposed to clear the default cache after syncing location data? Similar to how we clear incompelte session data

app/src/main/res/values/strings.xml
JohnnieShen: If trying to sync without internet or sync itself fails, I can probably split into 2 if you want it

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: Traditionally it's a mixed bag, ImagingViewModel uses lateinit but RegistrationViewModel uses injection

app/src/main/java/com/vci/vectorcamapp/core/data/mappers/FormQuestionMapper.kt
JohnnieShen: Because i needed to use it to cast prerequisite DTO into domains in the method

aryamanshodhan: I did not have to do that when pulling in the registration screen though? Can you use a similar architecture to how it is in the registration view model?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
aryamanshodhan: Lets make this consistent. Can you look up what the expected pattern is? Let's stick to that. Create a task to fix any other screens.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
aryamanshodhan: Oh I see. Yeah we should clear out the default cache.

app/src/main/res/values/strings.xml
aryamanshodhan: I'd say we should have 2 different messgaes, one for no internet, and one for other failure.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: Think injection is the better pattern, i have changed it and can add a task

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsScreen.kt
```diff
@@ -68,12 +73,75 @@ fun SettingsScreen(
...
                 )
+
                 SettingsActionTile(
                     title = "Start Practice Session",
                     onClick = { onAction(SettingsAction.StartNewPracticeSession) },
                     modifier = modifier
                 )
             }
+
+            SettingsSection(title = "Data Synchronization") {
```
aryamanshodhan: Code here is good. I would consider changing the text throughout this component to something that is a little bit less jargony and more accessible for people who might not have a tech background. Also, would you say this is considered "syncing" data? We arent really sending data to the backend here no?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -28,26 +46,46 @@ import javax.inject.Inject
 
 @HiltViewModel
 class SettingsViewModel @Inject constructor(
+    private val transactionHelper: TransactionHelper,
```
aryamanshodhan, line 50: Subsequent PR: We should make it consistent throughout the codebase, whether we should inject here or as a lateinit var. Create a JIRA task for this.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +259,141 @@ class SettingsViewModel @Inject constructor(
                         emitError(SettingsError.COLLECTOR_DELETION_FAILED)
                     }
                 }
+                SettingsAction.ResyncProgramData -> {
```
aryamanshodhan, line 279: Add a line between actions

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +259,141 @@ class SettingsViewModel @Inject constructor(
...
+                    resyncData()
+                }
+            }
+        }
+    }
+
+    private suspend fun resyncData() {
+        _state.update { it.copy(isSyncingData = true) }
+        try {
+            val program = state.value.program
```
aryamanshodhan: Replace with _state.value.program

Also, create a JIRA task to replace all instances of state.value in all view models to _state.value

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +259,141 @@ class SettingsViewModel @Inject constructor(
...
+            val programId = program.id
+
+            val incompleteSessions = sessionRepository.observeIncompleteSessionsAndSites().first()
```
aryamanshodhan: incompleteSessionsAndSites. Question: Do we have a get version for getting this data? Or do we need to observe? I'd recommend using get, not observe.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +259,141 @@ class SettingsViewModel @Inject constructor(
...
+            val currentSession = currentSessionCache.getSession()
+            val defaultFields = defaultIntakeFieldsCache.getDefaultIntakeFields()
+
+            val success = transactionHelper.runAsTransaction {
+                fetchAndSeedAllLocationTypesForProgram(programId)
+                fetchAndSeedAllSitesForProgram(programId)
+                fetchAndSeedFormForProgram(programId)
+
+                incompleteSessions.forEach { sessionAndSite ->
```
aryamanshodhan: Great catch! This is very important. Does this guarantee that in every possible user flow, the session will need to select the site again? AKA is there a scenario where we might be saving and trying to upload a session with site ID = -1?

aryamanshodhan: Another thing you need to do for each of the incomplete sessions is delete all the formAnswers for each of the sessions. Because there might be new form questions

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +259,141 @@ class SettingsViewModel @Inject constructor(
...
+                        collectorTitle = defaultFields.collectorTitle,
+                        collectorLastTrainedOn = defaultFields.collectorLastTrainedOn,
+                        hardwareId = defaultFields.hardwareId,
+                        district = "",
+                        villageName = ""
+                    )
+                }
+
+                if (currentSession != null) {
+                    currentSessionCache.saveSession(currentSession, siteId = -1)
```
aryamanshodhan: Same thing for the current session:

Does this guarantee that in every possible user flow, the session will need to select the site again? AKA is there a scenario where we might be saving and trying to upload a session with site ID = -1?

Another thing you need to do for each of the incomplete sessions is delete all the formAnswers for each of the sessions. Because there might be new form questions

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +259,141 @@ class SettingsViewModel @Inject constructor(
...
+        questionDtos.forEach { questionDto ->
+            formQuestionRepository.upsertFormQuestion(questionDto.toDomain(), formId, parentId).onError { error ->
+                emitError(error)
+                throw Exception("Failed to save surveillance form questions")
+            }
+            questionDto.subQuestions?.let { subQuestions ->
+                seedFormQuestionsForProgram(subQuestions, formId, questionDto.id)
             }
         }
     }
```
aryamanshodhan, line 418: Question: Did you make that one public function private again? I forgot the name.

aryamanshodhan (commented):
Well done Johnny! Home stretch. Just a couple of final comments.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: I think it should cover all scenerios

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: Yeah i did

app/src/main/java/com/vci/vectorcamapp/settings/domain/util/SettingsError.kt
```diff
@@ -4,5 +4,7 @@ import com.vci.vectorcamapp.core.domain.util.Error
 
 enum class SettingsError : Error {
     COLLECTOR_SAVE_FAILED,
-    COLLECTOR_DELETION_FAILED
+    COLLECTOR_DELETION_FAILED,
+    DATA_SYNC_FAILED,
+    DATA_SYNC_IN_PROGRESS_SESSION_EXIST
```
aryamanshodhan, line 9: Question: What does the error on line 9 mean?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +275,148 @@ class SettingsViewModel @Inject constructor(
...
+        _state.update { it.copy(isSyncingData = true) }
+        try {
+            val program = _state.value.program
+            val programId = program.id
+
+            val incompleteSessions = sessionRepository.getIncompleteSessionsAndSites()
+            val currentSession = currentSessionCache.getSession()
+
+            if (incompleteSessions.isNotEmpty() || currentSession != null) {
+                emitError(SettingsError.DATA_SYNC_IN_PROGRESS_SESSION_EXIST)
```
aryamanshodhan, line 296: This is okay as a temporary solution. Eventually we should migrate to having a situation where we display a warning if the user is using an outdated form or site.

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +275,148 @@ class SettingsViewModel @Inject constructor(
...
+
+            val success = transactionHelper.runAsTransaction {
+                fetchAndSeedAllLocationTypesForProgram(programId)
+
+                siteRepository.setAllSitesInactiveForProgram(programId)
+
+                fetchAndSeedAllSitesForProgram(programId)
+                fetchAndSeedFormForProgram(programId)
+
+                true
```
aryamanshodhan, line 311: Question: If any of the fetches or seeds fail, does the code still fall through and set success to true?

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -221,6 +275,148 @@ class SettingsViewModel @Inject constructor(
...
+                        collectorName = defaultFields.collectorName,
+                        collectorTitle = defaultFields.collectorTitle,
+                        collectorLastTrainedOn = defaultFields.collectorLastTrainedOn,
+                        hardwareId = defaultFields.hardwareId,
+                        district = "",
+                        villageName = ""
+                    )
+                }
+
+                if (currentSession != null) {
```
aryamanshodhan: Shouldn't we remove this now? We cannot set siteId to -1...if the database for this session is not updated, we should certainly not set the cache to -1...otherwise they will be out of sync from each other and might cause unexpected side effects.

aryamanshodhan (changes requested):
Hey Johnny! This is good, but there is a potential bug/unintended side effect. Please resolve it.

app/src/main/java/com/vci/vectorcamapp/settings/domain/util/SettingsError.kt
JohnnieShen: If you try to sync while there is still incomplete sessions

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: Pretty sure it doesn't since we are using the transation helper, if any exception thrown the entire transaction gets cancelled and success will not be set to true

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
```diff
@@ -322,10 +322,6 @@ class SettingsViewModel @Inject constructor(
                         villageName = ""
                     )
                 }
-
-                if (currentSession != null) {
```
aryamanshodhan, line 326: Question: Is there a scenario where currentSession will be non-null while on this settings screen? If so, might need to set delete currentSession from the cache. Can you double check where currentSessionCache is saved and deleted? My thought is, it is non-null during the intake and imaging screens, and then the landing screen. But just double check this.

aryamanshodhan (approved):

app/src/main/java/com/vci/vectorcamapp/settings/presentation/SettingsViewModel.kt
JohnnieShen: It's possible for currentSession to be non null while on settings because it is not cleared anywhere. Auto deleting here might not be a good idea neither

PR #203, opened 2026-04-15
SOF-462: Allow decimal point answers in numerical DynamicFormField

ThomasT0830:
This pull request improves number input validation and user experience in the form components, and adjusts UI layout in the registration screen. The main changes are stricter validation for numeric answers, enhanced input filtering for number fields, and a UI tweak for consistent button height.

**Form validation and input handling:**

* Updated `ValidateFormAnswersUseCase` to enforce stricter validation for "number" type answers: now allows decimals, but rejects values that start or end with a dot, or have more than two decimal places.
* Improved the number input field in `DynamicFormField` to only allow numbers with up to two decimal places, and prevents invalid input as the user types.

**UI improvements:**

* Set a consistent medium height for the confirm program button in the registration screen for better UI alignment.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateFormAnswersUseCase.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateFormAnswersUseCase.kt
@@ -28,7 +28,7 @@ class ValidateFormAnswersUseCase @Inject constructor() {
 
                 else -> when (question.type) {
                     "number" -> {
-                        if (answer.toDoubleOrNull() == null || answer.startsWith(".") || answer.endsWith(".") || answer.substringAfter(".", "").length > 2) {
+                        if (answer.toDoubleOrNull() == null || answer.startsWith(".") || answer.endsWith(".")) {
                             Result.Error(FormValidationError.INVALID_FORM_ANSWER)
                         } else Result.Success(Unit)
                     }
--- a/app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateFormAnswersUseCase.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateFormAnswersUseCase.kt
@@ -27,9 +27,11 @@ class ValidateFormAnswersUseCase @Inject constructor() {
                 answer.isBlank() -> Result.Success(Unit)
 
                 else -> when (question.type) {
-                    "number" -> if (answer.toIntOrNull() == null) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
```diff
@@ -39,7 +39,11 @@ fun DynamicFormField(
             TextEntryField(
                 label = question.label,
                 value = value,
-                onValueChange = { newValue -> onValueChange(newValue.filter { it.isDigit() }) },
+                onValueChange = { newValue ->
+                    if (newValue.isEmpty() || newValue.matches(Regex("^\\d+\\.?\\d{0,2}$"))) {
```
aryamanshodhan, line 43: Explain this regex expression please.

app/src/main/java/com/vci/vectorcamapp/intake/domain/use_cases/ValidateFormAnswersUseCase.kt
```diff
@@ -27,9 +27,11 @@ class ValidateFormAnswersUseCase @Inject constructor() {
                 answer.isBlank() -> Result.Success(Unit)
 
                 else -> when (question.type) {
-                    "number" -> if (answer.toIntOrNull() == null) {
-                        Result.Error(FormValidationError.INVALID_FORM_ANSWER)
-                    } else Result.Success(Unit)
+                    "number" -> {
+                        if (answer.toDoubleOrNull() == null || answer.startsWith(".") || answer.endsWith(".") || answer.substringAfter(".", "").length > 2) {
```
aryamanshodhan: Maybe make a new validation error, because saying "Invalid answer" for >2 decimal places is misleading. Alternatively, restrict typing only up to 2 decimal places after a "." is typed.

aryamanshodhan (changes requested):
One comment Thomas. Looks good otherwise!

app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
ThomasT0830: `\d+` means that there must be a positive number of digits before the decimal if there is a decimal, or a positive number of digits

`.?` means that there may or may not be a decimal

`\d{0,2}` means that there is either 0, 1, or 2 digits after the decimal (we have to allow 0 digits after the decimal in the UI layer because it is a valid intermediate level)

`^` and `$` just mark the start and end of the string, respectively

aryamanshodhan (approved):

PR #209, opened 2026-04-21
SOF-467: Create label for specimen + confidence with the bounding box overlay in the imaging screen

JohnnieShen:
This pull request refactors the `BoundingBoxOverlay` composable to improve how bounding boxes and their labels are rendered on the camera preview. The main change is switching from drawing with a `Canvas` to using `Box` layouts, which allows for easier positioning and styling of the bounding box and its associated confidence label. Additionally, a new padding dimension is introduced for finer control of label spacing.

Bounding box rendering improvements:
* Refactored the `BoundingBoxOverlay` composable to use `Box` layouts instead of `Canvas`, enabling easier overlay of text labels and bounding boxes, and simplifying the calculation of positions and dimensions using Compose's layout system. [[1]](diffhunk://#diff-51a7013c6a3dabf44339d18bb224570d3efa755063de7144f48f2ba7812f37bcL3-R13) [[2]](diffhunk://#diff-51a7013c6a3dabf44339d18bb224570d3efa755063de7144f48f2ba7812f37bcL23-R59)
* Added a confidence label above each bounding box, displaying the specimen confidence value with improved styling and positioning, including background color and padding.

UI consistency:
* Introduced a new `paddingExtraExtraSmall` value in the `Dimensions` class for more granular control of padding, used in the label styling.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
@@ -1,12 +1,16 @@
 package com.vci.vectorcamapp.imaging.presentation.components.camera
 
-import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.geometry.Offset
-import androidx.compose.ui.geometry.Size
-import androidx.compose.ui.graphics.drawscope.Stroke
+import androidx.compose.ui.layout.layout
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.unit.IntSize
 import com.vci.vectorcamapp.core.domain.model.InferenceResult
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
```diff
@@ -20,25 +24,39 @@ fun BoundingBoxOverlay(
     modifier: Modifier = Modifier
 ) {
     val density = LocalDensity.current
-    val strokeWidth = with(density) { MaterialTheme.dimensions.borderThicknessThick.toPx() }
-
+    val strokeWidth = MaterialTheme.dimensions.borderThicknessThick
```
aryamanshodhan, line 27: Question: Why did we remove the with (density) part?

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
```diff
@@ -20,25 +24,39 @@ fun BoundingBoxOverlay(
...
+
+        Box(
+            modifier = Modifier
+                .size(width = widthDp, height = heightDp)
+                .border(width = strokeWidth, color = boxColor)
         )
     }
-}
\ No newline at end of file
+}
```
aryamanshodhan, line 68: Question: Do we not need to draw the rectangle using a canvas? Just double check that the rectangle dimensions are not off for any reason...Absolute and relative dimensions might play a picture when you add the box. Is the box's dimensions relative to the camera view or relative to the immediate parent? Just try out multiple cases and mosquito positions to confirm this please.

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
JohnnieShen: Because the old code uses native Ui instead of compose, it is no longer needed

app/src/main/java/com/vci/vectorcamapp/imaging/presentation/components/camera/BoundingBoxOverlay.kt
JohnnieShen: Dimensions I can check but this is swapping from using native UI to compose UI. Not sure why we used native Ui in the first place

aryamanshodhan (approved):
Approving on the condition that bounding box is unchanged. Please do a final round of testing. Also add to excel sheet.

PR #219, opened 2026-04-27
feat(SOF-465): add * for mandatory fields

stephen1706:
add * for mandatory fields in form screen

Screenshot_20260427-110251

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
@@ -23,10 +23,11 @@ fun DynamicFormField(
     modifier: Modifier = Modifier,
     error: Error? = null,
 ) {
+    val label = if (question.required) "${question.label} *" else question.label
     when (question.type) {
         "text" -> {
             TextEntryField(
-                label = question.label,
+                label = label,
                 value = value,
                 onValueChange = onValueChange,
                 singleLine = true,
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
aryamanshodhan, line 1: Question: Can we make the (*) red in color?

aryamanshodhan (changes requested):

app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
stephen1706: edited

app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
```diff
@@ -23,10 +23,12 @@ fun DynamicFormField(
     modifier: Modifier = Modifier,
     error: Error? = null,
 ) {
+    val label = question.label
```
aryamanshodhan, line 26: Optional: Can we inline this wherever we say label = label?

aryamanshodhan (dismissed):
Looks good Stephen! Just a final optional comment.

app/src/main/java/com/vci/vectorcamapp/intake/presentation/components/DynamicFormField.kt
stephen1706: it is not possible, because DatePickerField have selectedDateInMillis as first param, if we dont set label = label, it will be compilation error

aryamanshodhan (approved):

PR #223, opened 2026-06-05
feat: add detekt check

stephen1706:


aryamanshodhan (commented):
Looks good, but just confirming this is config for detekt. Until we fix all files in the project according to Detekt linting, we will not have passing CI/CD?

aryamanshodhan (dismissed):

aryamanshodhan (approved):

PR #230, opened 2026-07-09
feat: create background service during upload image

stephen1706:
## Summary
- Adds `UploadForegroundService`, a `START_STICKY` foreground service that anchors the app process for the entire duration of an upload — from the moment work is enqueued through completion of both the metadata and image upload workers.
- Tags both `MetadataUploadWorker` and `ImageUploadWorker` with `session_upload_work` so the service can observe all active upload work via `WorkManager.getWorkInfosByTagFlow()`.
- Injects `@ApplicationContext` into `WorkManagerRepositoryImplementation` and calls `UploadForegroundService.start()` immediately after enqueuing work, closing the process-kill window that existed while workers were in the ENQUEUED state.
## Why
WorkManager's `SystemForegroundService` only runs as a foreground service once `doWork()` begins. While work is **ENQUEUED** (e.g. waiting for network), there is no foreground service — the OS is free to kill the process if the app is in the background. Additionally, `SystemForegroundService` is not `START_STICKY`, so Android will not restart it if it is terminated mid-upload.
`UploadForegroundService` solves both problems:
1. Starts a foreground service **before** workers run (eliminating the ENQUEUED window).
2. Returns `START_STICKY` so the OS recreates and restarts the service if it is killed, then re-attaches to in-flight WorkManager tasks.
3. Declared with `android:stopWithTask="false"` so it continues running even after the user swipes the app away from recents.
4. Self-terminates by observing WorkManager state and calling `stopSelf()` once no upload work remains active.

### DIFF
--- a/app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
+++ b/app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
@@ -24,6 +24,8 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 import javax.inject.Inject
 
+private enum class UploadState { WAITING, UPLOADING, IDLE }
+
 /**
  * Sticky foreground service (START_STICKY) that anchors the app process during active uploads.
  *
@@ -43,6 +45,9 @@ class UploadForegroundService : Service() {
 
     private val serviceScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
     private var observerJob: Job? = null
+    private val notificationManager by lazy {
+        getSystemService(NOTIFICATION_SERVICE) as NotificationManager
+    }
 
     companion object {
         const val UPLOAD_WORK_TAG = "session_upload_work"
@@ -65,13 +70,14 @@ class UploadForegroundService : Service() {
     }
 
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
-        startForeground(NOTIFICATION_ID, buildNotification(), ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC)
+        startForeground(NOTIFICATION_ID, buildNotification("Preparing upload\u2026"), ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC)
         startObservingWork()
         return START_STICKY
     }
 
     /**
-     * Observes WorkManager upload tags and stops the service once all upload work is settled.
+     * Observes WorkManager upload tags, updates the notification text to reflect the current
+     * state, and stops the service once all upload work is settled.
      *
      * Uses a `seenActiveWork` flag so we do not stop prematurely if the initial observation
      * happens before WorkManager has registered the enqueued tasks, and to correctly handle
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
```diff
@@ -0,0 +1,137 @@
...
+ * WorkManager's SystemForegroundService only starts a foreground service once doWork() begins,
+ * leaving a window where the process can be killed while work is still ENQUEUED. This service
+ * fills that gap and also ensures Android restarts the process if the service is killed mid-upload.
+ *
+ * Lifecycle:
+ *  - Started by WorkManagerRepository immediately before upload work is enqueued.
+ *  - Observes WorkManager state and calls stopSelf() when no upload work remains active.
+ *  - On system restart (START_STICKY), re-attaches to any in-flight WorkManager tasks.
+ */
+@AndroidEntryPoint
```
aryamanshodhan, line 41: Can you explain the difference between @AndroidEntryPoint and just @Inject constructor()?

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    private val serviceScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+    private var observerJob: Job? = null
+
+    companion object {
+        const val UPLOAD_WORK_TAG = "session_upload_work"
+
+        private const val NOTIFICATION_ID = 9001
+        private const val CHANNEL_ID = "upload_foreground_service_channel"
+        private const val CHANNEL_NAME = "Upload Service"
```
aryamanshodhan, line 57: Will this create separate notifications for each session that is being uploaded?

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
```diff
@@ -0,0 +1,137 @@
...
+                    it.state == WorkInfo.State.RUNNING ||
+                    it.state == WorkInfo.State.ENQUEUED ||
+                    it.state == WorkInfo.State.BLOCKED
+                }
+
+                when {
+                    hasActiveWork -> seenActiveWork = true
+                    seenActiveWork -> stopSelf() // active → idle transition: all done
+                    workInfos.isNotEmpty() -> stopSelf() // all work already completed/failed/cancelled
+                    // workInfos.isEmpty() → work not yet enqueued, keep waiting
```
aryamanshodhan: Do we need this state? It seems commented right now.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    override fun onBind(intent: Intent?): IBinder? = null
+
+    override fun onDestroy() {
+        super.onDestroy()
+        serviceScope.cancel()
+    }
+
+    private fun buildNotification(): Notification {
```
aryamanshodhan: If we are building the notification and creating the channel here, do we need to create it again in the work manager repository implementation/metadata upload worker/image upload worker?

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
aryamanshodhan, line 1: I think with the addition of this file, there might now be some dead work manager code. Would you be able to do a second pass through all data upload related files to see if any code is now dead?

aryamanshodhan (changes requested):
Looks great Stephen! Can you just look through the existing upload related files to see if there is any dead code?

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
stephen1706: -- | @Inject constructor | @AndroidEntryPoint
-- | --|--
Who creates the object | Hilt | The Android OS
Injection style | Constructor | Fields (@Inject lateinit var)
Used on | Repositories, Workers, Use cases, etc. | Activity, Fragment, Service, BroadcastReceiver
Hilt's role | Build + inject the whole object | Inject fields into an already-created object

stephen1706: basically android entry point is to tell hilt when the app is first launch, we will initialize this class, while inject is the typical dependency injection to inject a field

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
stephen1706: because the channel id is the same, it will be always be 1 notification showing that to tell user there is a bckground service running, this is needed otherwise the OS will kill the service when app is in background. 

The per-session notifications come from the workers, which were already doing this before in `MetadataUploadWorker`, this class is still needed to show the progress of uploading and its not a dead code. basically this UploadForegroundService is to allow `MetadataUploadWorker` run in background

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
stephen1706: its not needed because the service will just keep running and do nothing when there is nothing enqueue on work manager. the commented code is just to make it clearer.

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
stephen1706: there is no dead code because the existing work manager is still needed to upload the api, this service is just to allow our app run in backround (but need to show need push notification due to OS requirement)

app/src/main/java/com/vci/vectorcamapp/core/data/upload/UploadForegroundService.kt
stephen1706: yes there will be 2 separate notification (as seen on screenshot), the MetadataUploadWorker will manage the upload progress, while this class just show `Uploading session data` push notification to allow our app run in background

ThomasT0830 (approved):
Looks good to me!

aryamanshodhan (approved):

PR #231, opened 2026-07-16
Migrate inference to LiteRT 2.1.6 CompiledModel with GPU

stephen1706:
## Summary
- Upgrade on-device ML runtime from TFLite/LiteRT Interpreter to **LiteRT 2.1.6** `CompiledModel` API
- Enable **GPU acceleration** (`Accelerator.GPU`) with automatic CPU fallback for detector and classifiers
- Resolve per-model input shapes from buffer requirements (e.g. species 512×512 vs sex/abdomen 300×300) and switch logging to Timber

## Test plan
- [ ] Fresh install/build; confirm models initialize without buffer-size errors
- [ ] Logcat shows GPU accelerator when supported, or CPU fallback when not
- [ ] Detection still returns valid boxes
- [ ] Species / sex / abdomen classification still produce logits
- [ ] Close imaging session and confirm models/buffers release cleanly


ThomasT0830 (dismissed):
Looks good to me!

aryamanshodhan (approved):

